PR #26, opened 2018-06-07
We should always lowercase true and false when converting JSON

Yasumoto:
My calls to [`DescribeAutoScalingGroups`](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DescribeAutoScalingGroups.html) returned a value of `False`. This caused a crash via `JSONDecoder`, which could not handle the input (and was expecting a lowercased `false`).

This change now defensively handles the case where we receive a capitalized boolean value back from the API.


noppoMan:
Grate catching! LGTM. 
@jonnymacs Do you have any opinions? If not, I'll merge this.

Yasumoto:
Thanks folks! 🙇

PR #28, opened 2018-07-07
Remove "Members" key when parsing lists from AWS response

### LINKED ISSUE #27: `DescribeAutoScalingGroupsResult` contains a member field
With this code:

```
import AWSSDKSwift

let shape = Autoscaling.AutoScalingGroupNamesType(maxRecords: nil, nextToken: nil, autoScalingGroupNames: ["fake-asg"])
let autoscaling = Autoscaling()
let response = try autoscaling.describeAutoScalingGroups(shape)

for group in response.autoScalingGroups {
    group.autoScalingGroupName
}
```

We get an error trying to parse out `AvailabilityZones`:

```
.build/x86_64-unknown-linux/debug/Jaguar2
Fatal error: Error raised at top level: Swift.DecodingError.keyNotFound(CodingKeys(stringValue: "AvailabilityZones", intValue: nil), Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "AutoScalingGroups", intValue: nil), _DictionaryKey(stringValue: "Index 0", intValue: 0)], debugDescription: "No value associated with key CodingKeys(stringValue: \"AvailabilityZones\", intValue: nil) (\"AvailabilityZones\").", underlyingError: nil)): file /home/buildnode/jenkins/workspace/oss-swift-4.1-package-linux-ubuntu-16_04/swift/stdlib/public/core/ErrorType.swift, line 191
```

This is because [`DescribeAutoScalingGroups`](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DescribeAutoScalingGroups.html) has a response structured like this:

```
<DescribeAutoScalingGroupsResponse xmlns="http://autoscaling.amazonaws.com/doc/2011-01-01/">
  <DescribeAutoScalingGroupsResult>
    <AutoScalingGroups>
      <member>
        <AvailabilityZones>
...
```

(with an extra `<member>`) as opposed to:

```
<DescribeAutoScalingGroupsResponse xmlns="http://autoscaling.amazonaws.com/doc/2011-01-01/">
  <DescribeAutoScalingGroupsResult>
    <AutoScalingGroups>
      <AvailabilityZones>
...
```

It looks like we need to handle this nested `member` inside of `Instances` and a few other fields as well.

Yasumoto:
This fixes #27.

When AWS sends a list as a response, it typically wraps it as `{"Members": [A, B, C]}`. This breaks our JSON decoding, since our objects do not expect to look up a `Members` key, and just find the values of `[A, B, C]` directly.

We are now iterating through dicts -and lists to dive into each sub-structure, and look for any instance of `Members`, removing it. I've added a few test cases for parsing the dictionaries, which were a great start.

However, I found that pulling in the example cases from #27 directly served as quite helpful test cases, hence the extra `Model` files I'm including. @noppoMan and @jonnymacs, what do you two think of this copy-pasting from `aws-sdk-swift`? It's easy to remove, but I kind of like the extra coverage. Happy either way!

Also, if anyone has suggestions for ways to improve the algorithm, please let me know!


Yasumoto:
This is based off https://github.com/noppoMan/aws-sdk-swift-core/pull/29, so that should probably be merged first.

.gitignore
```diff
@@ -2,3 +2,4 @@
 /.build
 /Packages
 /*.xcodeproj
+*.swp
```
jonnymacs: curious what brings about this local swp file?

Yasumoto: @jonnymacs using `vim` 😅

Yasumoto:
@jonnymacs did ya have a chance to take a look at how the Ruby SDK handles this?

Looks [like the Go SDK does indeed need to handle "nested shapes"](https://github.com/aws/aws-sdk-go/blob/v1.15.1/private/model/api/shape.go#L569) and [parse 'em out](https://github.com/aws/aws-sdk-go/blob/v1.15.1/private/model/api/api.go#L768).

Yasumoto:
We're going to go with #30 instead 💪

Issue #15, opened 2017-12-07
issue with tls ciphers when used with vapor


jonnymacs:
Hi Yuki,

I ran into a problem that I was able to work around but I'd like to better understand the cause to see if the workaround is viable or if I'll need another solution.

I built a package that includes swift-aws/kinesis and implements a simple interface to put records in one of the various kinesis streams I have setup.

I am then adding the package to a vapor project, which is an http api service that will take a request, and put the contents in a kinesis stream.

This is working on OS X. However upon building a docker image - ubuntu 16.04, swift 4.0.2, and vapor 3.1 I saw an error - `Fatal error: Unexpectedly found nil while unwrapping an Optional value` when calling putRecord on a Kinesis instance.

Debugging, I was able to trace the issue to the clibressl swift package, specifically this function call to set the tls ciphers https://github.com/vapor-community/clibressl/blob/23ddb296981d17a8ee6c7418742a40cad5d2f9d0/Sources/CLibreSSL/ssl_lib.c#L1292 was returning 0 instead of 1.

My ssl expertise is limited, and I am not sure why, but it is only when executed in the vapor app that this occurs. Running the swift REPL in the docker container, and importing CLibreSSL, I cannot reproduce. 

My workaround is to add CLibreSSL as an explicit dependency to my vapor package manifest. However I am a bit uncomfortable not knowing exactly what the problem is and / or consequences of doing this. It seems to me that CLibreSSL is no longer being maintained in favor of vapor/tls, but again I'm not sure what if anything should be done.  

I do think it is important that this sdk work with vapor without too many additional config changes as it's an increasingly popular framework for building swift services.

Please let me know your thoughts?

Thanks again for this great library!

noppoMan:
@jonnymacs 
I'll try to investigate this issue on ubuntu 16.04.
Could you send me the content of Package.swift?

jonnymacs:
ok, i'm learning more as I go here.

In an effort to simplify the repro steps I included the swift-aws/kinesis module in my vapor package manifest, but after doing that I started seeing compilation errors about duplicate symbols between CHTTP (which is included with the vapor/engine) and CHTTParser that is included with aws-sdk-swift-core. It seems to me that the http_parser in these two libs is identical... So now wondering if you might consider using the vapor/engine in the Prorsum package... (CHTTParser is now in the ZewoGraveyard, so it likely will not see any updates moving fwd.)

[...]

PR #29, opened 2018-07-07
Migrate package to Swift 4, remove support for Swift 3

Yasumoto:
This also modifies the package dependencies to use a slightly upgraded `HypertextApplicationLanguage`, but we can remove that if https://github.com/noppoMan/HypertextApplicationLanguage/pull/1 gets merged.


Yasumoto:
@noppoMan How do you feel about bumping the version of Swift to 4.1? I think there are a few good improvements here, and we'll need it for #28 and #15 🙏

jonnymacs:
@Yasumoto lgtm. I think Yuki was working on this for the parent sdk as well.

@noppoMan i will go ahead and merge this tomorrow morning if no objections

https://github.com/noppoMan/aws-sdk-swift/pull/75

Yasumoto:
Sweet! Sounds like I could lend a hand on some build issue there, I’ll take a look tomorrow (I’m in Dublin, so it’s UTC for me) 👍

jonnymacs:
@Yasumoto ok, what comes to mind is whether this warrants a version bump... since there is impact on the parent sdk, which still claims support for 3.x

Perhaps we should update the parent first?

PR #30, opened 2018-08-02
serialize elements to json that are called "Member" with special logic

jonnymacs:
Do not create "member" key in dictionary structure

### DIFF
--- a/Tests/AWSSDKSwiftCoreTests/SerializersTests.swift
+++ b/Tests/AWSSDKSwiftCoreTests/SerializersTests.swift
@@ -11,65 +11,101 @@ import XCTest
 @testable import AWSSDKSwiftCore
 
 class SerializersTests: XCTestCase {
-    
+
     struct B: AWSShape {
         public static var members: [AWSShapeMember] = [
             AWSShapeMember(label: "a", required: true, type: .string),
             AWSShapeMember(label: "b", required: false, type: .list),
             AWSShapeMember(label: "c", required: true, type: .list)
         ]
-        
+
         let a = "1"
         let b = [1, 2]
         let c = ["key": "value"]
     }
-    
+
     struct C: AWSShape {
         public static var members: [AWSShapeMember] = [
             AWSShapeMember(label: "value", required: true, type: .string)
         ]
-        
+
         let value = "hello"
     }
-    
+
     struct D: AWSShape {
         public static var members: [AWSShapeMember] = [
             AWSShapeMember(label: "value", required: true, type: .string)
         ]
-        
+
         let value = "world"
     }
-    
+
+    struct E: AWSShape {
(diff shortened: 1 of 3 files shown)


Tests/AWSSDKSwiftCoreTests/SerializersTests.swift
```diff
@@ -11,65 +11,101 @@ import XCTest
...
-    
+
     func testSerializeToXML() {
         let node = try! AWSShapeEncoder().encodeToXMLNode(A(), attributes: ["A": ["url": "https://example.com"]])
         XCTAssertEqual(node.attributes["url"], "https://example.com")
-        
+
         let xml = XMLNodeSerializer(node: node).serializeToXML()
-        let expected = "<A url=\"https://example.com\"><Structure><A>1</A><B>1</B><B>2</B><C><key>value</key></C></Structure><DList><Value>world</Value></DList><CList><Value>hello</Value></CList><Array>foo</Array><Array>bar</Array></A>"
+        let expected = "<A url=\"https://example.com\"><Structure><A>1</A><B>1</B><B>2</B><C><key>value</key></C></Structure><DList><Value>world</Value></DList><CList><Value>hello</Value></CList><Array>foo</Array><Array>bar</Array><StructureWithMember><Member><memberKey2>memberValue2</memberKey2><memberKey>memberValue</memberKey></Member></StructureWithMember><StructureWithMembers><Member><memberKey2>memberValue2</memberKey2><memberKey>memberValue</memberKey></Member><Member><memberKey2>memberValue2</memberKey2><memberKey>memberValue</memberKey></Member></StructureWithMembers></A>"
```
Yasumoto: Interestingly, [I found the GO SDK to use a similar looking test](https://github.com/aws/aws-sdk-go/blob/master/models/protocol_tests/output/ec2.json#L161) and uses this syntax: `<member>abc</member>`

Yasumoto: Yep, and [they're looking for the `member` too](https://github.com/aws/aws-sdk-go/blob/6a946b29774a7b037a96feb5ea6da48b5df35702/private/protocol/xml/xmlutil/unmarshal.go#L139).

Yasumoto:
This looks awesome, and will close #27.

I tested with this:

```
let shape = Iam.ListServerCertificatesRequest()
let iam = Iam()
let response = try iam.listServerCertificates(shape)
print(response)
```

which is broken without this change. Thanks! 🎉

Tests/AWSSDKSwiftCoreTests/SerializersTests.swift
Yasumoto: If you want to, it should be `member` and not `Member`, but that's just a nit.

PR #31, opened 2018-08-16
replace Prorsum with swift-nio

jonnymacs:


noppoMan:
@jonnymacs @Yasumoto 
Thats' Awesome!!! Great job you guys!
I didn't see the detail yet, but I'm interested in how to handle networking with swift-nio with synchronous syntax.

PR #35, opened 2018-09-02
Add shared credentials support

oliveroneill:
I've added INIParser as a dependency for reading the config files.

I also removed the SharedCredential.default property since I was unsure what it was intended for. Instead when AWSClient is created, it will attempt to create a SharedCredentials object.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -11,6 +11,15 @@
         }
       },
       {
+        "package": "INIParser",
+        "repositoryURL": "https://github.com/PerfectlySoft/Perfect-INIParser.git",
+        "state": {
+          "branch": null,
+          "revision": "0952aac9ca54324ff25191569fd9c48eec424772",
+          "version": "3.0.2"
+        }
+      },
+      {
         "package": "swift-nio",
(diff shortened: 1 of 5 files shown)


Yasumoto:
Sweet! Thank you for getting this added!

On first glance, this looks like a pretty clean/simple dependency to add.

Any chance you could add a test for the newly implemented code you added?

oliveroneill:
No problem. Just committed new tests.

Yasumoto:
Thank you @oliveroneill!

@noppoMan I think it's alright to bring it in. @jonnymacs mentioned that since we have the SwiftJSON parser from Kitura, bringing in small-scoped dependencies for targeted usecases is reasonable.

If it turns out that we need to fork that code and bring it over, we can move it under the `swift-aws` org. What do y'all think?

Yasumoto:
I'm going to go ahead and merge this in 🎉 

@noppoMan if I misread you, please let me know and I'm happy to put in the effort to either fork this inside the `swift-aws` organization or help with other tweaks.

@oliveroneill thank you again for contributing this!! 🙇

PR #36, opened 2018-09-06
Improve XML list parsing

Yasumoto:
When calling ELB's `describe-load-balancers`, I found we were parsing the XML incorrectly, creating JSON like this:

```
"BackendServerDescriptions":[{"InstancePort":8080,"PolicyNames":{}},{"InstancePort":8443,"PolicyNames":{}}]
```

Converting that into the respective shape didn't work as expected, since we're somehow turning `<PolicyNames><member>TFEnableProxyProtocol</member></PolicyNames>` into `"PolicyNames":{}`. 

This is [specified correctly](https://github.com/aws/aws-sdk-go/blob/master/models/apis/elasticloadbalancing/2012-06-01/api-2.json#L1357) in our JSON definition, so it's definitely how we're converting it.

I've just submitted the below test case in the hopes that someone else knows how to fix this while I sleep on it 🤔

### DIFF
--- a/Sources/AWSSDKSwiftCore/Serializer/XMLNodeSerializer.swift
+++ b/Sources/AWSSDKSwiftCore/Serializer/XMLNodeSerializer.swift
@@ -122,10 +122,16 @@ public class XMLNodeSerializer {
             let memberChildren: [XMLNode]
 
             if arrayNodes.isEmpty {
-                jsonStr += "{"
-                memberChildren = memberNode.children.filter({ !keys.contains($0.elementName) })
-                jsonStr += _serialize(nodeTree: memberChildren)
-                jsonStr += "}"
+                if memberNode.children.isEmpty {
+                    jsonStr += "["
+                    jsonStr.append(contentsOf: memberNode.values.flatMap(formatAsJSONValue))
+                    jsonStr += "]"
+                } else {
+                    jsonStr += "{"
+                    memberChildren = memberNode.children.filter({ !keys.contains($0.elementName) })
+                    jsonStr += _serialize(nodeTree: memberChildren)
+                    jsonStr += "}"
+                }
             } else {
                 memberChildren = node.children.filter({ !keys.contains($0.elementName) })
                 for (_, nodes) in arrayNodes {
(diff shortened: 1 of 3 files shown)


jonnymacs:
ok I can get you 

```
{"BackendServerDescriptions":[{"InstancePort":80,"PolicyNames":"TFEnableProxyProtocol"} ...
```

without much effort. But making `PolicyNames` a list is going to be a challenge. In fact it's clear now we don't reference the datatype from the json def when we parse at all, but rather rely on xml conventions. A single value is a scalar, not an array.

Assuming `PolicyNames` is a list I'll need a little more time to get this right.

Yasumoto:
In retrospect, this should've been more clear to me based on a few things. I wrote out my test string on paper to see the hierarchy:

fullsizeoutput_96fe

I realized that we needed to find the place where we were taking care of nodes with three properties:

1. The `XMLNode`'s `elementName` is `Member`
2. There are 0 children nodes
3. The `value` attached to the node is of type `[String]`.

Not sure how legit this is or not, I'm running some real-world tests to see how this goes.

Sources/AWSSDKSwiftCore/Serializer/XMLNodeSerializer.swift
```diff
@@ -122,10 +122,16 @@ public class XMLNodeSerializer {
             let memberChildren: [XMLNode]
 
             if arrayNodes.isEmpty {
-                jsonStr += "{"
-                memberChildren = memberNode.children.filter({ !keys.contains($0.elementName) })
-                jsonStr += _serialize(nodeTree: memberChildren)
-                jsonStr += "}"
+                if memberNode.children.isEmpty {
```
jonnymacs: brilliant. I was up in the `_serialize` method trying to analyze memberNode (with `.hasSingleValue()` for example) to skip jumping into `_processNodeWithMember` entirely, but this is much simpler and more elegant

Yasumoto: 🙌 

This also makes sense as the place to edit since the previous output was getting us `{}`, showing this is definitely the right spot to be putting in the conditional.

Yasumoto: Yup, this gets me further along 👍

PR #37, opened 2018-09-19
URLEncode the equals sign _just_ in the value of a query string

Yasumoto:
I was trying to [upload a certificate](https://docs.aws.amazon.com/IAM/latest/APIReference/API_UploadServerCertificate.html) which included a `=` and it was failing AWS' parsing.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -315,12 +315,12 @@ extension AWSClient {
 
             for key in keys {
                 if let value = dict[key] {
-                    queryItems.append("\(key)=\(value)")
+                    let stringValue = "\(value)"
+                    queryItems.append("\(key)=\(stringValue.replacingOccurrences(of: "=", with: "%3D"))")
                 }
             }
 
             if let params = queryItems.joined(separator: "&").addingPercentEncoding(withAllowedCharacters: CharacterSet(charactersIn: String.uriAWSQueryAllowed.joined())) {
-
                 if path.contains("?") {
                     path += "&" + params
                 } else {
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -314,11 +314,14 @@ extension AWSClient {
             dict["Action"] = operationName
             dict["Version"] = apiVersion
 
-            if let urlEncodedQueryParams = urlEncodeQueryParams(fromDictionary: dict) {
-                body = .text(urlEncodedQueryParams)
+            switch httpMethod {
+            case "GET":
+                queryParams = queryParams.merging(dict) { $1 }
+            default:
+                if let urlEncodedQueryParams = urlEncodeQueryParams(fromDictionary: dict) {
+                    body = .text(urlEncodedQueryParams)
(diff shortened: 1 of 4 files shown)


jonnymacs:
any luck implementing the generic encoder?

jonnymacs:
I can take a stab at it if that helps. Trying to get this merged, then we can tag rc-2.

Yasumoto:
If I don't get a diff posted tonight, please feel free to run with it! 🙇

jonnymacs:
@Yasumoto let me know your thoughts 

A few notes:

To add tests for createAWSRequest I added an extension to build a private method that can be imported for testing in debug modes. Let me know if you hate this idea.

[URLQueryItem](https://developer.apple.com/documentation/foundation/urlqueryitem) does not seem to encode forward slashes that are included in query string values. I have to assume this is ok. 

`value=<html><a = http://foo.com></a></html>` is encoded as `value=%3Chtml%3E%3Ca%20%3D%20http://foo.com%3E%3C/a%3E%3C/html%3E`

jonnymacs:
phooey - need to tell travis to ack the preprocessor directive...

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -307,28 +310,15 @@ extension AWSClient {
...
-
-                if path.contains("?") {
-                    path += "&" + params
-                } else {
-                    path += "?" + params
-                }
-
-                body = .text(params)
-            }
+            queryParams = queryParams.merging(dict) { $1 }
```
jonnymacs: is there ever a case where the Input.queryParams share a key with what is in the input and this merge will overwrite a key?

jonnymacs:
@Yasumoto this week maybe?

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -368,7 +364,37 @@ extension AWSClient {
...
+            if let value = dict[key] {
+                queryItems.append(URLQueryItem(name: key, value: String(describing: value)))
+            }
+        }
+        return queryItems.isEmpty ? nil : queryItems
+    }
+}
+
+// debug request creator
+#if DEBUG
```
Yasumoto: whoa, nice

Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
```diff
@@ -0,0 +1,125 @@
...
+@testable import AWSSDKSwiftCore
+
+class AWSClientTests: XCTestCase {
+
+    struct C: AWSShape {
+        public static var members: [AWSShapeMember] = [
+            AWSShapeMember(label: "value", required: true, type: .string)
+        ]
+
+        let value = "<html><body><a href=\"https://redsox.com\">Test</a></body></html>"
```
Yasumoto: ⚾️

Sources/AWSSDKSwiftCore/AWSClient.swift
jonnymacs: thnx.  was actually worried this might be poorly received :)

jonnymacs:
@Yasumoto how are we looking? I hate to merge without you're testing it out

PR #44, opened 2018-12-26
Invoke async

jonnymacs:
Refactor of AWSClient to return Swift Nio EventLoopFutures.

This will require a change to all the API signatures, as well as documentation.

I decided to put this out there for review . We are using NIO for http and calling wait() makes me uncomfortable, even if clients can use a DispatchQueue to make it work. And it's not clear how long until swift5 async / await is ready / implemented.

In this PR we also run createNioRequest async in case the MetaDataService has to run getCredentials.

With this change, the interface will look something like this, and can be safely run from the main thread.

```swift
import S3 
import NIO

do {
    let bucket = "my-bucket"

    let s3 = S3(
        accessKeyId: "Your-Access-Key",
        secretAccessKey: "Your-Secret-Key",
        region: .uswest2
    )

    // Create Bucket, Put an Object, Get the Object
    let createBucketRequest = S3.CreateBucketRequest(bucket: bucket)

    try s3.createBucket(createBucketRequest).thenThrowing { response -> Future<S3.PutObjectOutput> in
        // Upload text file to the s3
        let bodyData = "hello world".data(using: .utf8)!
        let putObjectRequest = S3.PutObjectRequest(acl: .publicRead, bucket: bucket, contentLength: Int64(bodyData.count), body: bodyData, key: "hello.txt")
        return try s3.putObject(putObjectRequest)
    }.thenThrowing { response -> Future<S3.GetObjectOutput> in
        let getObjectRequest = S3.GetObjectRequest(bucket: bucket, key: "hello.txt")
        return try s3.getObject(getObjectRequest)
    }.whenSuccess { futureResponse in
        futureResponse.whenSuccess { response in
          if let body = response.body {
              print(String(data: body, encoding: .utf8))
          }
        }
    }
} catch {
    print(error)
}
```

@Yasumoto @noppoMan what are your thoughts?


noppoMan:
@jonnymacs 
I mostly agree with this changes for now.  this was only a matter of time when we decided to adopt swift-nio.

Actually I don't know the detail of current execution model of `AWSSDKSwift`(sorry..), So I need to take a look at both of it and this PR changes. So could you give me some time? I'll be able to reply early in the new year.

jonnymacs:
@noppoMan Happy New Year 🎉 

Sounds good. 👍 

The change required in all of the service apis is straight forward.  import NIO in the api files, then methods with output return Future<T>.  

I have a branch in the main project `return_futures` that is using this branch of sdk-core  you can try it when you get a chance.

Here is an example: 

```
// THIS FILE IS AUTOMATICALLY GENERATED by https://github.com/noppoMan/aws-sdk-swift/blob/master/Sources/CodeGenerator/main.swift. DO NOT EDIT.

import Foundation
import AWSSDKSwiftCore
import NIO

import S3Middleware

/**

*/
public struct S3 {

    let client: AWSClient

    public init(accessKeyId: String? = nil, secretAccessKey: String? = nil, region: AWSSDKSwiftCore.Region? = nil, endpoint: String? = nil) {
        self.client = AWSClient(
            accessKeyId: accessKeyId,
            secretAccessKey: secretAccessKey,
            region: region,
            service: "s3",
            serviceProtocol: ServiceProtocol(type: .restxml),
            apiVersion: "2006-03-01",
            endpoint: endpoint,
            serviceEndpoints: ["us-west-2": "s3.us-west-2.amazonaws.com", "eu-west-1": "s3.eu-west-1.amazonaws.com", "us-east-1": "s3.amazonaws.com", "ap-northeast-1": "s3.ap-northeast-1.amazonaws.com", "s3-external-1": "s3-external-1.amazonaws.com", "ap-southeast-2": "s3.ap-southeast-2.amazonaws.com", "sa-east-1": "s3.sa-east-1.amazonaws.com", "ap-southeast-1": "s3.ap-southeast-1.amazonaws.com", "us-west-1": "s3.us-west-1.amazonaws.com"],
            partitionEndpoint: "us-east-1",
            middlewares: [S3RequestMiddleware()],
            possibleErrorTypes: [S3ErrorType.self]
        )
    }

    ///  Creates a new bucket.
    public func createBucket(_ input: CreateBucketRequest) throws -> Future<CreateBucketOutput> {
        return try client.send(operation: "CreateBucket", path: "/{Bucket}", httpMethod: "PUT", input: input)
    }

    // etc...
}
```

adam-fowler:
Just had a quick look at this. Looking good. I'm building an NIO2.0 branch off it. I noticed if a call returns Future<Void> this isn't passed onto the user so they can't wait on the operation completing.

adam-fowler:
Also because you cannot wait on Future< Void > you cannot catch any errors returned.

adam-fowler:
In fact looking at the code aws-sdk-swift has never checked for errors being returned on commands that don't return a value. Maybe we should fix that

adam-fowler:
@jonnymacs I'm ready for it to go in. I've been using it for the last 3-4 weeks with no issues. As I mentioned above it would be good to return futures from all operations so we can check for errors but that can be done later.

jonnymacs:
@adam-fowler @Yasumoto sorry to tack onto this, but I wanted to fix an issue that has plagued me the past 2 days before merging and tagging.

HEAD and GET requests were failing from ecs containers that run with IAM role credentials. Other credential providers worked, but not when the credentials had a token.

2 issues I fixed - we were not including the token in the URL on GET requests, and we were using signedHeaders to issue HEAD requests. I have addressed both in the last few commits here and feel pretty confident it's ready.

If you want to take a look please do so, else I will move forward and merge this today and cut a tag

PR #46, opened 2019-02-08
properly serialize the payloadPath object as json / xml

jonnymacs:
if the shape has a payload path and it defines a shape, we need to properly serialize that shape.

unfortunately, when the protocol is JSON convincing the compiler that the reflection object is a generic AWSShape that's encodable is not as simple as a switch statement... so we need to serialize the full input to JSON, back to a dictionary, parse the payloadPath out and serialize that... a bit ugly, but effective.

Thank you to @adam-fowler for finding this bug and sharing the solution.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -301,7 +301,15 @@ extension AWSClient {
         switch serviceProtocol.type {
         case .json, .restjson:
             if let payload = Input.payloadPath, let payloadBody = mirror.getAttribute(forKey: payload.toSwiftVariableCase()) {
-                body = Body(anyValue: payloadBody)
+                switch payloadBody {
+                case is AWSShape:
+                    let inputBody: Body = .json(try AWSShapeEncoder().encodeToJSONUTF8Data(input))
+                    if let inputDict = try inputBody.asDictionary(), let payloadDict = inputDict[payload] {
+                        body = .json(try JSONSerialization.data(withJSONObject: payloadDict))
+                    }
+                default:
+                    body = Body(anyValue: payloadBody)
+                }
                 headers.removeValue(forKey: payload.toSwiftVariableCase())
             } else {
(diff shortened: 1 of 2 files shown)


jonnymacs:
@Yasumoto @noppoMan if you have time today / tomorrow, else I'll merge this

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -301,7 +301,15 @@ extension AWSClient {
         switch serviceProtocol.type {
         case .json, .restjson:
             if let payload = Input.payloadPath, let payloadBody = mirror.getAttribute(forKey: payload.toSwiftVariableCase()) {
-                body = Body(anyValue: payloadBody)
+                switch payloadBody {
+                case is AWSShape:
+                    let inputBody: Body = .json(try AWSShapeEncoder().encodeToJSONUTF8Data(input))
```
Yasumoto: Ouch, yah this is ugly. I think this is just showing we need to have better data-transformation options available here, likely through `Codable` and proper encoders/decoders of the various formats, yah?

jonnymacs: This one is really tough b/c the object we need to encode is actually a variable on the Input instance who's name is dynamically defined by the `payload` string. To access this, @noppoMan implemented reflection with the Mirror class, but this returns an object of type `Any`, so `Codable` won't work. The compiler simply won't allow us to pass the object to the encoder unless it knows for sure that the dataType will always be a `Codable` object (which is doesn't since the type is `Any`).

Even if we add control flow that only calls the encoder if `payloadBody` is an `AWSShape`, it doesn't work b/c it's type is still `Any` as far as the compiler is concerned. 

Upcasting to an `AWSShape` doesn't work, b/c the encoder wants a Generic type as input, not a protocol.

So I don't have a much better solution that this really. Perhaps some super class that implements `AWSShape`, then every service shape extends this class... but seems like a lot to change to resolve this...

Issue #48, opened 2019-02-28
Shared Credentials override supplied credentials


adam-fowler:
If I have shared credentials on my systems, they are always used regardless of whether I supply credentials to the AWSClient or not. I would have thought it should be the other way where they are used if I don't supply credentials to the AWSClient.

jonnymacs:
@adam-fowler agreed. order should be

`static, env, shared, iam`

Yasumoto:
This is fixed thanks to #51, right?

PR #50, opened 2019-03-18
JSONSerializer and Dictionaries fix

adam-fowler:
serializeToFlatDictionary() was not serializing dictionaries correctly. They were being serialized as follows
```
<dictionary name>.<key name>=value
```
When they should have been serialized as follows
```
<dictionary name>.entry.N.key=<key name>
<dictionary name>.entry.N.value=<value>
```
where N is the index of the key,entry pair

I also added code to fix the shared credentials fix from Issue #48


adam-fowler:
Might want to take this one back, given the way the flatten dictionary test is failing, but the issue of serialising dictionaries still exists. Looks like this might produce unwanted results for structures within structures. Needs another review

adam-fowler:
DItching this as there doesn't seem to be anyway of telling the difference between a structure and dictionary at this point even though AWS is expecting them in a different format. Going to try creating a DictionaryShape that outputs the correct format

PR #54, opened 2019-04-05
Add Support for EventLoopFuture

AAAstorga:
Added support for futures so we don't block the thread. I also have a PR to implement the rest of these changes in the other repo.


jonnymacs:
@AAAstorga this is great. thanks for submitting this PR.

Not sure if you noticed there is an existing PR for this change #44 

Please take a look. The change you made is pretty much the same as what's in #44, but the bigger difference in #44 is that in addition to the send methods there is an http call in the MetaDataService when credentials are retrieved via IAM on ec2 instances and ecs containers. So createNioRequest needs to be executed async `.wait()` will still get called on the main thread.

The code changes in #44 account for this.

There is a branch https://github.com/swift-aws/aws-sdk-swift/tree/return_futures that was generated and also includes updates to documentation. Considering this is a rather significant change it is planned for a 3.0 release.

Assuming the changes in this pr are working for you, that's great. But if you want to take a look at #44 and offer some feedback or testing that would be really helpful. Also it's been awhile since that branch (and the return_futures branch) were rebased off their respective masters if you want to take a stab at that.

PR #55, opened 2019-04-14
Swift 5 + NIO 2

foulkesjohn:
Use BoringSSL for HMAC + SHA256 functions.


adam-fowler:
Why don't you use the BoringSSL versions of the HMAC, SHA256 functions which are included in NIO 2.0 instead of including another version of OpenSSL?

foulkesjohn:
@adam-fowler didn't think of that. Will update 👍

foulkesjohn:
Updated. I've had to put a few more expected values in for the xml serialisation, not sure if this is the best way but I managed not to get failing tests after a few runs.

jonnymacs:
@foulkesjohn yeah, i have some code in the async branch to make that serialization test more dynamic and easier to maintain.

jonnymacs:
@foulkesjohn @adam-fowler just so I understand, with this change we will no longer support swift4? What about compatibility with vapor3?

Yasumoto:
@jonnymacs maybe we do one release with #44 and #53, and tag that `2.0.0` (I don't see a non-rc version for `2.0.0`, right?) so we support Swift 4 & Vapor 3, then we create a branch off that tag to support that version.

Then we can merge this and support Swift 5 & NIO 2 on master with `3.0.0`?

Alternatively, this diff isn't _too_ bad, and we *could* support both for a bit like [micro-express does](https://github.com/NozeIO/MicroExpress/blob/0.5.2/Sources/MicroExpress/ServerResponse.swift#L25).

adam-fowler:
> @foulkesjohn @adam-fowler just so I understand, with this change we will no longer support swift4? What about compatibility with vapor3?

This might not help just now, but vapor has moved to Swift5 and NIO 2 on their master branch. They just haven't done a release yet.

Yasumoto:
Thought this over, and I think if we can wrap support for both swift4 and swift5 that’d be non-breaking and support folks for a while.

Since Vapor 4 doesn’t even have an alpha yet, I don’t think we want to divide our for a few more months. However I personally would love to simplify and use NIO2/swift5 with this project and not worry about merging in a branch for a long haul.

Totally open to suggestions tho!

adam-fowler:
Hi John, Have you got this code to run ok? I have an almost identical set of changes and while they compile fine the code gets stuck in future.wait() as soon as I try to connect with AWS. After much delving into the NIO code it seems the SSL handshake is failing.

foulkesjohn:
I have it running. The thing that caught me out was migrating the handlers over and renaming ctx to context. There is a migration  here: https://github.com/apple/swift-nio/blob/master/docs/migration-guide-NIO1-to-NIO2.md. Double check everything against this and the api changes.

On 22 Apr 2019, at 01:44, Adam Fowler <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi John, Have you got this code to run ok? I have an almost identical set of changes and while they compile fine the code gets stuck in future.wait() as soon as I try to connect with AWS. After much delving into the NIO code it seems the SSL handshake is failing.

adam-fowler:
Ah missed the context bit, sorted now

adam-fowler:
@Yasumoto @jonnymacs Until we have this worked out, do you guys have any issue with me creating nio2swift5 branches in aws-sdk-swift and aws-sdk-swift-core? Any work required to support both versions of swift can be done on these branches.

jonnymacs:
That sounds like a good idea. I will say I’d prefer to merge the
invokeAsync / return_futures branches first and call it 2.1. This is a
pretty big change to all the interfaces, and a good amount of the core
code. So it would be best to branch from there.

Hoping to do this in the coming week. I’m running the invoke asynchronous
branch in production on 10% traffic with no issues so far.

adam-fowler:
Sounds good. Wish the async code was available when I first came to this, possibly would have structured my own stuff differently. 

Although I would add a load of documentation for this, as the async code assumes a prior knowledge of swift nio.

foulkesjohn:
Just hit an issue when running this code. Looks like assertions have been added to `wait()` which will throw when called inside an event loop: 

So this: https://github.com/swift-aws/aws-sdk-swift-core/pull/44 probably needs to go with this PR.

adam-fowler:
@foulkesjohn have you merged in the invokeAsync branch? 

I have added two branches to the aws-sdk-swift repositories which combine both the nio2.0 code and the invokeAsync code. nio2.0-invokeAsync in aws-sdk-swift-core and nio2.0-return_futures in aws-sdk-swift. Can you tell me if these work for you? If so can we close this pull request.

foulkesjohn:
Just give it a try and it seems to be fixed 👍

adam-fowler:
@foulkesjohn can we close this pull request?

foulkesjohn:
@adam-fowler yep 👍

foulkesjohn:
@adam-fowler where has the nio2.0-invokeAsync branch gone? I was using that until the final release is done. Do you have another branch I can use in the meantime?

adam-fowler:
@foulkesjohn sorry it got so messed up with merges I rebuilt it. There is a nio2.0 branch which is much cleaner

PR #57, opened 2019-04-25
Resurrect custom endpoint request

noppoMan:
Hi, I'm back.
Today I tried to request dynamodb-local through custom endpoint like...
```
DynamoDB(endpoint: "http://localhost:8000")
```
But current AWSClient seems to lost the function of custom endpoint.
(`MetaDataService.getCredential()` is always called and attempt to connect `169.254.x.x` when the credential is empty.)
So I fixed it to connect the custom endpoint when it's given.

### DIFF
--- a/Sources/AWSSDKSwiftCore/Signers/V4.swift
+++ b/Sources/AWSSDKSwiftCore/Signers/V4.swift
@@ -42,7 +42,7 @@ extension Signers {
             return sha256(data).hexdigest()
         }
 
-        public init(credential: CredentialProvider, region: Region, service: String, endpoint: String?) {
+        public init(credential: CredentialProvider, region: Region, service: String, endpoint: String? = nil) {
             self.credential = credential
             self.region = region
             self.service = service
--- a/Sources/AWSSDKSwiftCore/Signers/V4.swift
+++ b/Sources/AWSSDKSwiftCore/Signers/V4.swift
@@ -16,6 +16,8 @@ extension Signers {
         public let region: Region
 
         public let service: String
+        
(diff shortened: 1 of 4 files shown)


Sources/AWSSDKSwiftCore/Signers/V4.swift
```diff
@@ -40,21 +42,22 @@ extension Signers {
...
             self.service = service
+            self.endpoint = endpoint
         }
 
         public func signedURL(url: URL, date: Date = Date(), expires: Int = 86400) -> URL {
             let datetime = V4.timestamp(date)
             let headers = ["Host": url.hostWithPort!]
             let bodyDigest = hexEncodedBodyHash(Data())
-            let credentialForSignature = getCredential()
+            resolveCredentialIfNeeded()
```
jonnymacs: Hey @noppoMan! welcome back!

This looks great. 

The reason for assigning the signer credential to a `credentialForSignature` constant here is due to the fact that a concurrent request could change the Signer credential at the moment when credentials expire and rotate on an instance / container setup. So I think we should preserve getCredential returning the current CredentialProvider if that makes sense?

noppoMan:
https://github.com/swift-aws/aws-sdk-swift/issues/118

jonnymacs:
@noppoMan eager to get this merged, did you happen to see my comment? We need to make a copy of the credential instance var instead of relying on the instance var since it can change curing the signing process.

jonnymacs:
@noppoMan great thanks. yes I agree. I renamed to `manageCredential` in the async branch which will be the bases for a 2.1 release.

PR #58, opened 2019-04-30
Incorrect stringToSign created for API calls that include query parameters without assignation

adam-fowler:
If you have a path that includes a query parameter without an "=" sign the stringToSign that is built during the V4 signature construction is wrong. The code was outputting the query parameter without the "=" sign. AWS expects the "=" to be there. 

It seems that this only affects a number of S3 calls and a couple of calls in Cloudfront. I have checked both of systems and the code fixes the issue. This fixes https://github.com/swift-aws/aws-sdk-swift/issues/117

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -365,7 +365,11 @@ extension AWSClient {
         // add new queries to query item array. These need to be added to the queryItems list instead of the queryParams dictionary as added nil items to a dictionary doesn't add a value.
         if let pathQueryItems = parsedPath.queryItems {
             for item in pathQueryItems {
-                queryItems.append(URLQueryItem(name:item.name, value:item.value))
+                if let value = item.value {
+                    queryItems.append(URLQueryItem(name:item.name, value:value))
+                } else {
+                    queryItems.append(URLQueryItem(name:item.name, value:""))
+                }
             }
         }
 
@@ -407,7 +411,7 @@ extension AWSClient {
             if let value = dict[key] {
                 queryItems.append(URLQueryItem(name: key, value: String(describing: value)))
             } else {
-                queryItems.append(URLQueryItem(name: key, value: nil))
+                queryItems.append(URLQueryItem(name: key, value: ""))
             }
         }
         return queryItems.isEmpty ? nil : queryItems
--- a/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
+++ b/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
@@ -30,7 +30,7 @@ class AWSClientTests: XCTestCase {
             AWSShapeMember(label: "Member", required: true, type: .list),
         ]
 
-        let Member = ["memberKey": "memberValue", "memberKey2": "memberValue2"]
+        let Member = ["memberKey": "memberValue", "memberKey2" : "memberValue2"]
 
         private enum CodingKeys: String, CodingKey {
             case Member = "Member"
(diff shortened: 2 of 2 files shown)


jonnymacs:
@adam-fowler nice catch.

I worry a little about changing the url in the signer because `urlEncodedQueryParams` are sent in the body when the http method is not a GET, and they won't match what's in the signature.

It's also sad to have to perform another operation on the url in the signer...

I think this can be fixed in `createAWSRequest` without requiring another operation on the url string in the signer. If the `URLQueryItem` has a nil value, then setting the item name to `attr=` with value `nil` looks like it will produce the expected result.

Are you open to trying this?

`--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -401,11 +401,15 @@ extension AWSClient {
 
         // construct query array
         var queryItems = urlQueryItems(fromDictionary: queryParams) ?? []
-        
+
         // add new queries to query item array. These need to be added to the queryItems list instead of the queryParams dictionary as added nil items to a dictionary doesn't add a value.
         if let pathQueryItems = parsedPath.queryItems {
             for item in pathQueryItems {
-                queryItems.append(URLQueryItem(name:item.name, value:item.value))
+                if let value = item.value {
+                    queryItems.append(URLQueryItem(name: item.name, value: value))
+                } else {
+                    queryItems.append(URLQueryItem(name: "\(item.name)=", value: nil))
+                }
             }
         }
 
@@ -447,7 +451,7 @@ extension AWSClient {
             if let value = dict[key] {
                 queryItems.append(URLQueryItem(name: key, value: String(describing: value)))
             } else {
-                queryItems.append(URLQueryItem(name: key, value: nil))
+                queryItems.append(URLQueryItem(name: "\(key)=", value: nil))`

jonnymacs:
@adam-fowler nevermind.

jonnymacs:
value needs to be an empty string, try this I meant

```
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -401,11 +401,15 @@ extension AWSClient {
 
         // construct query array
         var queryItems = urlQueryItems(fromDictionary: queryParams) ?? []
-        
+
         // add new queries to query item array. These need to be added to the queryItems list instead of the queryParams dictionary as added nil items to a dictionary doesn't add a value.
         if let pathQueryItems = parsedPath.queryItems {
             for item in pathQueryItems {
-                queryItems.append(URLQueryItem(name:item.name, value:item.value))
+                if let value = item.value {
+                    queryItems.append(URLQueryItem(name: item.name, value: value))
+                } else {
+                    queryItems.append(URLQueryItem(name: "\(item.name)=", value: ""))
+                }
             }
         }
 
@@ -447,7 +451,7 @@ extension AWSClient {
             if let value = dict[key] {
                 queryItems.append(URLQueryItem(name: key, value: String(describing: value)))
             } else {
-                queryItems.append(URLQueryItem(name: key, value: nil))
+                queryItems.append(URLQueryItem(name: "\(key)=", value: ""))
             }
         }
```

adam-fowler:
Your code works as well, but for me it feels like you are massaging the inputs to a system to get it to work instead of fixing the system.

As for your concern about the ```urlEncodedQueryParams``` in the body for query based systems, the signature is only produced off the query params in the path which never get placed in the body. And as pointed out earlier this change only affects S3 and Cloudfront which use XML.

If you still prefer I'm happy to put that in though.

jonnymacs:
I see your point. 

But I'm still feeling pragmatic here. True the system doesn't support nil values for query params, and my proposed change does not "fix" that. But it seems like support for this minority case adds an excessive tax on the majority. I'll let you decide though, since you discovered this.

`urlEncodedQueryParams` are placed in the body when the serviceProtocol.type is `query` and the http method is not GET. Maybe it's not an issue though as you say.

Sources/AWSSDKSwiftCore/Signers/V4.swift
```diff
@@ -264,6 +264,17 @@ extension Signers {
             }
             return result
         }
+        
+        /// Encode Query according to https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
```
Yasumoto: Maybe add the relevant piece below that line too? (Up to you— it wasn't too hard to find but might be helpful)

```
`http://s3.amazonaws.com/examplebucket?acl`

The CanonicalQueryString in this case is as follows:
> `UriEncode("acl") + "=" + ""`
```

Sources/AWSSDKSwiftCore/Signers/V4.swift
```diff
@@ -10,17 +10,17 @@ import Foundation
 
 extension Signers {
     public final class V4 {
-
+
```
Yasumoto: Small nit, mind clearing this whitespace? 🙏

Yasumoto:
I think it sounds like there's a `QueryReformatter` which takes a good chunk of the urlencoding, query-rewriting, and assorted massaging away from both the `Client` and the `Signer`, tho that might be slightly longer term.

@adam-fowler do you know of a decent way to write a testcase to catch this as well? Just thoughtful of how we can catch any regressions going forward 🤔

adam-fowler:
@jonnymacs I'll go with your method. If I can think of a better way to support it within the V4 signature I might suggest a change later

@Yasumoto I can stick a test in the aws-sdk-swift S3 tests. If I go with Jonny's fix it's a lot harder to add a test at the core level, as I have to run the whole pipeline.

PR #62, opened 2019-05-06
fix HEAD requests

### LINKED ISSUE #61: HEAD requests do not work
NIO checks the header type to determine whether to parse a
body or not in the response. HEAD requests have no body.
    
Using the RAW header is an issue because the check in NIO
is restricted to request with method .HEAD.
RAW(value: "HEAD") == .HEAD is false.
    
see github.com/apple/swift-nio/blob/1.14.1/Sources/NIOHTTP1/HTTPDecoder.swift#L328

jonnymacs:
fixes #61

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -192,7 +192,7 @@ extension AWSClient {
         var nioRequest = nioRequest
 
         guard let url = URL(string: nioRequest.head.uri), let _ = url.hostWithPort else {
-            fatalError("nioRequest.head.uri is invalid.")
+            throw RequestError.invalidURL("nioRequest.head.uri is invalid.")
         }
 
         // TODO avoid copying
@@ -201,6 +201,9 @@ extension AWSClient {
             headers[key.description] = value
         }
 
+        // We need to convert from the NIO type to a string
+        // when we sign the headers and attach the auth cookies
+        //
         let method = "\(nioRequest.head.method)"
 
         let signedHeaders = signer.signedHeaders(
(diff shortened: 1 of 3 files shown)


adam-fowler:
Looks good. Any reason this is a fatalError over throwing a runtime error.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -190,27 +190,18 @@ extension AWSClient {
 
     func nioRequestWithSignedHeader(_ nioRequest: Request) throws -> Request {
         var nioRequest = nioRequest
+
+        guard let url = URL(string: nioRequest.head.uri), let _ = url.hostWithPort else {
+            fatalError("nioRequest.head.uri is invalid.")
```
Yasumoto: I think we can/should use [our already-defined](https://github.com/swift-aws/aws-sdk-swift-core/pull/62/files#diff-7863205517fdde3415e746c5fee58bc6R27) `RequestError.invalidURL` if possible?

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
```diff
@@ -93,13 +93,36 @@ public struct AWSRequest {
...
+          method: nioHTTPMethod(from: awsRequest.httpMethod),
+          uri: awsRequest.url.absoluteString
+        )
         let generatedHeaders = headers.map { ($0, $1) }
         head.headers = HTTPHeaders(generatedHeaders)
 
         return Request(head: head, body: try awsRequest.body.asData() ?? Data())
     }
 
+    fileprivate func nioHTTPMethod(from: String) -> HTTPMethod {
```
Yasumoto: Maybe add a note here

```
// We need to convert from the NIO type to a string when we sign the headers and attach the auth cookies.
```

since I took a bit to remember why the heck we were doing this anyway (and not just using NIO's builtin values instead of the stringly conversion.

Yasumoto:
Any chance you'd be up to add some test coverage? Maybe to `toNIORequest` ?

jonnymacs:
@adam-fowler @Yasumoto updated.

PR #63, opened 2019-05-06
Fail response future when connection fails

adam-fowler:
If there was no connection a noConnection error is thrown and the ClientBootstrap channel is never initialized. This means the response eventloopFuture is never connected up and the code will hang when you wait on it. 

To fix this I have added a thenIfErrorThrowing to catch the error and then set the response to fail inside there. 

Also added code to ensure the client.close() is always run

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -105,14 +105,16 @@ extension AWSClient {
         let client = HTTPClient(hostname: nioRequest.head.headers["Host"].first!, port: 443)
         let future = try client.connect(nioRequest)
 
-        //TODO(jmca) don't block
-        let response = try future.wait()
-
-        client.close { error in
-            if let error = error {
-                print("Error closing connection: \(error)")
+        future.whenComplete {
+            client.close { error in
+                if let error = error {
+                    print("Error closing connection: \(error)")
+                }
             }
         }
+
+        //TODO(jmca) don't block
+        let response = try future.wait()
 
         return response
     }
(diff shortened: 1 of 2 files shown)


Yasumoto:
Instead of using the `response` promise, what if we just return the future provided from bootstrapping the Channel directly? Then we get a bit simpler error/success handling and don't have the extra promise hanging around, right?

Yasumoto:
Asked in discord to [see if I'm wildly offbase](https://discordapp.com/channels/431917998102675485/448584561845338139/574990431973015553) too just in case.

adam-fowler:
Saw your conversation in discord. The response promise supplies the HTTP response. The future returned from the bootstrapping the channel returns void. Similarly cascade(to:) requries another Future< Void >

adam-fowler:
@Yasumoto So I dug into this a little more. Looked at various implementations of HttpClients (Amzn Smoke, Vapor). The Amzn client is a bit shit, it uses a callback to supply the response data. Vapor client does what we need, but is a little more complex. It will create a future from the ClientBootstrap that contains the response data. I'll tidy up the code and probably post a new pull request sometime in the next few days.

Yasumoto:
@adam-fowler before you go too far.. I think we should throw it out in favor of the SSWG-favored https://github.com/swift-server/swift-nio-http-client

Only catch is they haven’t tagged a version, so I was thinking of waiting until a release is cut before switching over. What do you think?

adam-fowler:
Looking through the code for that client their solution for the connect() failing is pretty much the same as I have in this change. They use whenFailure() instead of flatMapErrorThrowing which I can change to using.

adam-fowler:
damn seemed to have messed up this branch, bringing in unnecessary stuff.

adam-fowler:
So I changed to using whenFailure() like swift-nio-http-client. If we could at least push that through as that would resolve the problem I am seeing. I'll add an issue about investigating swift-nio-http-client as a possible replacement for the aws-sdk-swift http client.

Regarding versioning, we could fork it and add a version ourselves until they get around to versioning it.

Yasumoto:
(sorry for the speed bump with my comment btw— I totally forgot to all hit the `Approve` ✅ button! 😅 Definitely we need this merged pronto)

PR #68, opened 2019-05-13
Encoding/Decoding XML Requests and Responses using Codable

adam-fowler:
Not expecting this to go in now but wanted to put it up so you could have visibility on it.

I have added XML encoding/decoding of AWSShape using the swift codable system. This fixes multiple issues with S3 which uses XML plus a number of issues with systems that uses query requests but expect xml responses. There are a number of other minor fixes also included.

- When I added the query dictionary encoding I never set up EC2 to use it (can extract this into a separate pull request if required).
- Extended XML tests to cover more situations
- Added an XML namespace to AWSShapes. Some S3 requests fail if the namespace isn't provided
- Don't add an XML body, if there is no payload
- Sorry I seemed to have dragged in three changes from @noppoMan


jonnymacs:
@adam-fowler this all looks good. I have to say I don't have a good way to test it, other than it doesn't break anything I am using the sdk for (kinesis which is json, and simple s3 head / get requests)

So I guess I'm comfortable if you are. perhaps @Yasumoto has more regression cases to try?

adam-fowler:
@Yasumoto are you happy with this? I would like to pull it across when I get back to the UK tomorrow.

PR #71, opened 2019-05-18
return Future<Void> instead of Void

jonnymacs:
so clients can chain off the response and catch errors

forgive the lack of test, this is tough to test without calling out to aws...

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -146,8 +146,8 @@ extension AWSClient {
                 return try self.createNioRequest(awsRequest)
             }.then { nioRequest in
                 return self.invoke(nioRequest)
-            }.map { response in
-                return
+            }.thenThrowing { response in
+                return try self.validateCode(response: response)
             }
     }
 
@@ -162,8 +162,8 @@ extension AWSClient {
                 return try self.createNioRequest(awsRequest)
             }.then { nioRequest in
                 return self.invoke(nioRequest)
-            }.map { response in
-                return
+            }.thenThrowing { response in
+                return try self.validateCode(response: response)
             }
     }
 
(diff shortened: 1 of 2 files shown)


jonnymacs:
@jonnymacs @adam-fowler the more I think about it, it makes sense to get this in before cutting a 3.0 branch of the main sdk. This introduces an interface change and requires a rebuild of code and the docs. Since that's fresh in my scope right now I'll hold on cutting that tag until this gets a review.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -158,9 +160,11 @@ extension AWSClient {
                     httpMethod: httpMethod
                 )
                 return try self.createNioRequest(awsRequest)
-            }.whenSuccess { nioRequest in
-                _ = self.invoke(nioRequest)
-          }
+            }.then { nioRequest in
+                return self.invoke(nioRequest)
+            }.map { response in
```
Yasumoto: Why not return back the response instead of Void?

jonnymacs: hmm...

I thought about that. my reason not to was b/c the api spec says there is no output...

ex:

```
"AddTagsToCertificate":{
      "name":"AddTagsToCertificate",
      "http":{
        "method":"POST",
        "requestUri":"/"
      },
      "input":{"shape":"AddTagsToCertificateRequest"},
      "errors":[
        {"shape":"ResourceNotFoundException"},
        {"shape":"InvalidArnException"},
        {"shape":"InvalidTagException"},
        {"shape":"TooManyTagsException"}
      ]
    }
```

vs with an output

```
"DescribeCertificate":{
      "name":"DescribeCertificate",
      "http":{
        "method":"POST",
        "requestUri":"/"
      },
      "input":{"shape":"DescribeCertificateRequest"},
      "output":{"shape":"DescribeCertificateResponse"},
      "errors":[
        {"shape":"ResourceNotFoundException"},
        {"shape":"InvalidArnException"}
      ]
    }
```

So, if we return a NIO Response, are we conforming the the spec?

jonnymacs: oh and credit to @adam-fowler b/c he saw this and mentioned it in one of the other PRs :)

adam-fowler: @Yasumoto Surely, the only time you need the response is when there is an error. This can be dealt with in the same way as errors are dealt with for operations that return a non void response. See my comment above

jonnymacs:
@adam-fowler cool. just forced a push here - I missed passing `members:` into validateCode for the interfaces with output. but yeah, i will rebuild the code in the parent for this. I think tagging with @discardableResult makes sense as well.

PR #72, opened 2019-05-18
Nio2.0

adam-fowler:
Changes to get code running with swift-nio 2.0 and swift 5


adam-fowler:
@foulkesjohn

jonnymacs:
What are the requirements for nio2? swift5? will it run on ubuntu 16? These are some considerations. I recall an idea to maintain backward compatibility but if we can address with versioning then that's fine with me. Just need to make sure the Readme is clear on how to select a version.

Yasumoto:
Hm, good point. I was thinking we just do a clean-break, but we could use the [`swift()` compiler conditional](https://docs.swift.org/swift-book/ReferenceManual/Statements.html#grammar_compiler-control-statement) to use NIO2 only for > swift 5.

I think that’s a bit of extra maintenance burden, but maybe worth not needing to rev version numbers for a while?

adam-fowler:
@Yasumoto I've been running with this for the last month or so. It all seems in order. Moving to nio 2.0 generally meant a syntax change. There wasn't any functionality changes for us. There is an equivalent aws-sdk-swift branch nio2.0 which has a few swift 5 fixes in it.

I'm away for next 12 days so it is your call if you want it in now. And yes it would need a major version bump.

Yasumoto:
@adam-fowler totally makes sense! The only reason for the conditionals is we can support folks who haven’t upgraded to NIO2 yet without leaving them behind for a bit. I think it’s worth it to wrap these changes in the compiler conditionals for others who haven’t upgraded just yet... since you’re traveling would you be okay if someone merged those on top of your changes in this PR? (It doesn’t sound like there’s strong disagreement with maintaining backward compatibility, if there is then i’m In favor of just merging as-is!)

adam-fowler:
@Yasumoto I'm quite happy for someone else to take on the conditional work and merge it on top of my branch. 

Do we need to maintain swift 4.1 backwards compatibility as well?

adam-fowler:
> @adam-fowler totally makes sense! The only reason for the conditionals is we can support folks who haven’t upgraded to NIO2 yet without leaving them behind for a bit. I think it’s worth it to wrap these changes in the compiler conditionals for others who haven’t upgraded just yet... since you’re traveling would you be okay if someone merged those on top of your changes in this PR? (It doesn’t sound like there’s strong disagreement with maintaining backward compatibility, if there is then i’m In favor of just merging as-is!)

How would we optionally wrap the ```// swift-tools-version:5.0``` at the top of Package.swift? Are we just going to ask people to edit this locally? If so how will we get Travis to do a build testing the swift 4.2 build.

I can wrap the other changes in ```#if swift(>=5.0)```. There is a good chance we would have to repeat large blocks of code though. The nio EventLoopFuture calls .then() tend to chained together.

We could do something like this I guess
```
#if swift(<5.0)
extension EventLoopFuture {
    func flatMap<NewValue>(file: StaticString = #file, line: UInt = #line, _ callback: @escaping (T) -> EventLoopFuture<NewValue>) {
        then(file:file, line:line, callback)
    }
}
#endif
```
Really not keen on this though.

Personally I would prefer two branches.

Sources/AWSSDKSwiftCore/HMAC.swift
```diff
@@ -6,19 +6,18 @@
...
 //
 
-import Foundation
-import CNIOOpenSSL
+import CNIOBoringSSL
 
 func hmac(string: String, key: [UInt8]) -> [UInt8] {
     var context = HMAC_CTX()
-    HMAC_Init_ex(&context, key, Int32(key.count), EVP_sha256(), nil)
+    CNIOBoringSSL_HMAC_Init_ex(&context, key, key.count, CNIOBoringSSL_EVP_sha256(), nil)
```
kylebrowning: You probably don't want to use BOringSSL from NIO because its API will definitely be broken in the future. cc @weissi

adam-fowler: Hmmm, using the BoringSSL implementation means we don't need to link in another SSL lib. What are the plans that are going to break the API in the future?

kylebrowning: Im not sure, but most of the NIO team Ive talked to when they have reviewed my PRs, basically say 100% we shouldn't be using it. Heres one explanation. https://forums.swift.org/t/discussion-nioapns-nio-based-apple-push-notification-service/23384/11?u=kylebrowning

weissi: As @kylebrowning is pointing out, please definitely don't use NIO's internal modules. The only public API is what's exported by the [Package.swift](https://github.com/apple/swift-nio-ssl/blob/master/Package.swift#L30) as `.library` targets which for `swift-nio-ssl` is just `NIOSSL`.

That we are vending BoringSSL is literally just an implementation detail and the API does change a lot. We have (I think 3 times now) re-namespaced BoringSSL and BoringSSL itself _does not have_ a public API. So it's not us being unhelpful it's that BoringSSL doesn't have an API.

[BoringSSL](https://boringssl.googlesource.com/boringssl/) also documents that fact itself:

> Although BoringSSL is an open source project, it is not intended for general use, as OpenSSL is. We don't recommend that third parties depend upon it. Doing so is likely to be frustrating because there are no guarantees of API or ABI stability.
>
> Programs ship their own copies of BoringSSL when they use it and we update everything as needed when deciding to make API changes. This allows us to mostly avoid compromises in the name of compatibility. It works for us, but it may not work for you.

adam-fowler: ok, will look for an alternative

adam-fowler: @weissi are we ok to use swift-nio-ssl-support.git to load libressl, or should we just implement our own?

weissi: @adam-fowler yes, that's totally acceptable. Rule of thumb: Everything officially exported by a `Package.swift` is fine and `swift-nio-ssl-support` basically exports the system's libssl to you.

Andrea-Scuderi:
Hi @adam-fowler,
I've fixed this branch to work with swift 5.1. If you like I can add PR to the branch `nio2.0`

The solution is compatible with swift 5.0 as it uses `canImport(...)`.

https://github.com/Andrea-Scuderi/aws-sdk-swift-core/tree/nio2.0-swift5.1
and
https://github.com/Andrea-Scuderi/aws-sdk-swift/tree/nio2.0-swift5.1 (Cherry pick only the last commit)

adam-fowler:
Hi @Andrea-Scuderi, I’m currently on holiday. Is this something that’ll be needed on the master branch as well? If so could you make a pull request to there first.

PR #73, opened 2019-05-20
Only include a json body if there are members that should be included in that body

adam-fowler:
We are outputting a json body for operations where all the members are being placed elsewhere. These are creating operations with bodies where AWS isn't expecting one. This seems to screw up the signature generation. I think this is a similar problem to the XML body issue.

I've changed the code so if all the members are included in the url, header, or query params then don't output a json body.
 
This was found while trying to delete an archive in glacier.


adam-fowler:
As an aside, I had to make the assumption that shape members that didn't have a location were output in the body. Is there any reason we cannot default to this instead of having a nil location?

adam-fowler:
Also should we add a similar check for XML body output, ie if there are members that are output to the body then output a body.

adam-fowler:
@jonnymacs any thoughts on this

jonnymacs:
@adam-fowler took a closer look this morning. In general I am comfortable skipping the body encoding step if there is nothing to encode. However I am not sure the `bodyParams` dict is quite right (or necessary)? `bodyParams` will not necessarily be a `[String:String]`

So Kinesis for example, on many shapes the members do not have a location. And the output of bodyParams `["StreamName": "StreamName", "Records": "Records"]` for a `PutRecordsInput` is not "right".

If we want bodyParams to be and accurate dictionary representation of the body, I think we either need to do the horrible encodeToJson, decodeToDict...

Maybe though, all you want to do here is check "hasEncodableBody" or something to that affect that checks there are members with a body location or no location (which is basically what bodyParams does, I just don't like it's name)

adam-fowler:
Good point, I was trying to keep the interface consistent. But obviously the body elements can be more complex than a string. I had tried a hasEncodableBody() function previously will revert to that.

adam-fowler:
@jonnymacs what are your thoughts on the other two points I mention, regarding shape members that aren't given a location and doing the same for xml output instead of assuming a body isn't required.

jonnymacs:
@adam-fowler re:  shape members that aren't given a location

Yes, if there is no location, then the member goes in the body - as far as I can tell. So are you proposing to update AWSShapeMember to require location, then updating the code generator to set a location on all members? This seems fine but help me understand what we really gain from this?

And yes, makes sense to check if there is an encodable body for xml as well in all cases even when there is no payloadPath

adam-fowler:
@jonnymacs My suggestion for the location is a lot simpler. Don't make location optional and in the constructor for AWSShapeMember when setting the location do it as follows
```
self.location = location ?? .body(locationName: label)
```
We gain clarity ie it is obvious where a variable is going. We simplify code, don't need to check location is nil all the time. This isn't necessary, but just felt it was worth bringing up.

I'll add the XML check, and would you be happy for this to go in.

adam-fowler:
The XML check breaks S3.CreateBucket(). It adds a body for the xml. Then S3Middleware calculates the md5 for this body and then changes the contents of the body after.

There is a change on aws-sdk-swift to resolve this https://github.com/swift-aws/aws-sdk-swift/pull/129

adam-fowler:
@jonnymacs you happy for this one to go in now?

PR #75, opened 2019-05-28
Fix Swift 5 warnings

mxcl:


adam-fowler:
I thought I had seen those already go in,  obviously not. Do they break the swift 4.2 build? If not I would be happy to see them brought across. Although better wait to see if @jonnymacs or @Yasumoto have any issues. 

As an aside we are looking to bring across swift 5 and nio2.0 support in the near future, but also need to support swift 4.2 while Vapor doesn’t support swift 5.

mxcl:
It'll be safe with 4.2 since it just removes the duplication of accessor levels, which is what the warning is stating: that these are duplicate specifications.

mxcl:
Your CI is testing 4.2 as well of course.

jonnymacs:
@mxcl thanks for this PR. I am confused as well though, these changes are already in master. Am I missing something?

https://github.com/swift-aws/aws-sdk-swift-core/blob/master/Sources/AWSSDKSwiftCore/Decoder/DictionaryDecoder.swift#L1255

https://github.com/swift-aws/aws-sdk-swift-core/blob/master/Sources/AWSSDKSwiftCore/Decoder/DictionaryDecoder.swift#L1277

https://github.com/swift-aws/aws-sdk-swift-core/blob/master/Sources/AWSSDKSwiftCore/Decoder/UntypedDictionaryDecoder.swift#L1212

mxcl:
Oh my bad, didn’t check `master` first. Will close.

Look forward to a release!

PR #77, opened 2019-06-07
Nil payload

adam-fowler:
Output an empty body when there is payloadPath but the payloadBody is nil

Found another small issue, we were outputting the full body when a payloadPath pointed to an optional variable and that variable was nil.


jonnymacs:
this probably needs to be rebased after the encoder addition?

adam-fowler:
done

adam-fowler:
seems I missed the DictionaryDecoder changes, merged now

PR #83, opened 2019-06-10
Dictionary Encoder

adam-fowler:
- Based on the latest JSON encoder/decoder code.
- Use encoder to generate dictionary to extract json payload, instead of encoding JSON, decoding to a dictionary.


adam-fowler:
This also fixes https://github.com/swift-aws/aws-sdk-swift-core/issues/78

PR #84, opened 2019-06-14
Add percent encoded characters for urlEncodedQueryParams

adam-fowler:
The query string returned by AWSClient.urlEncodedQueryParams() did not percent encode enough characters. In particular it didn't encode the + sign. The function used URLComponents() to generate it's query params text. This would percent encode some characters but not all. I couldn't send it already percent encoded text as that percent encode the '%' signs. So instead I wrote a version which didn't use the URLComponents at all.

This fixes https://github.com/swift-aws/aws-sdk-swift/issues/107

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -442,13 +442,20 @@ extension AWSClient {
         )
     }
 
+    static let queryAllowedCharacters = CharacterSet(charactersIn:"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~:/")
     fileprivate func urlEncodeQueryParams(fromDictionary dict: [String:Any]) -> String? {
-        var components = URLComponents()
-        components.queryItems = urlQueryItems(fromDictionary: dict)
-        if components.queryItems != nil, let url = components.url {
-            return url.query
+        guard dict.count > 0 else {return nil}
+        var query = ""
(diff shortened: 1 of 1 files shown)


jonnymacs:
I'd very much prefer to use url components or some native lib to do this work if possible instead of maintaining a list inline. btw, where did you source the allowed chars list, I feel like I've seen many versions.  

A little confusing why `url.query` does not encoded everything correctly.  Have you looked at `urlComponents.setQueryItems(with: queryParams)`

I recall encoding the body on a query service was an issue for SES so we need to make sure everything is all good there if we make changes.

Last q, what are the conditions under which you have pre encoded data that is getting re-encoded?

adam-fowler:
I can understand why you'd like to use a native lib. But there doesn't seem to be a standard list of valid characters anywhere and these lists seem to change depending on what servers you are talking to. AWS servers don't like the '+' sign, so expect it to be percent encoded. This isn't particularly standard. AWS seems quite happy to decode/encode any character so not including a character in the list doesn't seem too problematic. I've tested all the special characters by passing them to a SQS queue and reading them all back. 

I can't find urlComponents.setQueryItems(with: queryParams).

The SES issue was a typo on my behalf in the AWSShapeEncoder code, not related to this.

To fix the issue I was having I had to encode the '+' sign. I couldn't encode it before setting the URLQueryItem() as it would encoded again, and I couldn't do the percent encoding of the final query string as it would encode the '=' and '&' characters. I guess I could do a percent encode of just the '+' sign of the final query string, but that doesn't feel much better than we current solution.

PR #85, opened 2019-06-14
Set EC2 content type to application/json

adam-fowler:
Even though EC2 outputs a querystring. The content-type should be set to application/json


adam-fowler:
Looks like I was mistaken, this isn't necessary

PR #86, opened 2019-06-15
Dictionary and Array encoding

adam-fowler:
For XML and query encoding there are multiple ways to encode dictionaries and arrays. This PR is an attempt to pull all of these together. The idea is to add an additional member to AWSShapeMember which defines how a collection is encoded. The collection encoding is defined by an enum with associated values

The options are as follows
- flatList: Arrays can be encoded without an enclosing element. Then the name of each array element is the locationName eg ```array.1=value1&array.2=value2```
- list(member): Arrays can be encoded with an enclosing element. Then the name of the enclosing element is the locationName and each array element gets it's name from the collection encoding. eg ```array.member.1=value1&array.member.2=value2```
- flatMap(key, value): Dictionaries can be encoded without an enclosing element. Then the name of each dictionary element is the locationName, and both the key element and value element get their name from the collection encoding.  eg ```dict.1.key=key1&dict.1.value=value1&dict.2.key=key2&dict.2.value=value2```
- map(entry, key, value): Dictionaries can be encoded with an enclosing element. Then the name of the enclosing element is the locationName. Each dictionary element, the key and value elements all get their names from the collection encoding eg ```dict.entry.1.key=key1&dict.entry.1.value=value1&dict.entry.2.key=key2&dict.entry.2.value=value2```.

The majority of the work after that was getting the XMLEncoder to work with this. I tried to keep the XML changes as separate from the rest of the code. It gets it's coding information from the AWSShapes via a protocol XMLContainerCodingMap.

Prior to this change there had been an effort to fix some of these issues via the CodeGenerator. Additional classes were created to wrap arrays if the member had a location name. This fixed the situation where you had a non-flattened array and needed the array element name. As these are not needed anymore I have removed them. This means some APIs will change slightly but for the better. eg
```
            let filters = [EC2.Filter(name: "group-name", values: ["test-group"])]
```
instead of
```
            let filters = EC2.FilterList(filter: [EC2.Filter(name: "group-name", values: EC2.ValueStringList(item: ["test-group"]))])
```

This change fixes https://github.com/swift-aws/aws-sdk-swift-core/issues/79 https://github.com/swift-aws/aws-sdk-swift-core/issues/80 https://github.com/swift-aws/aws-sdk-swift-core/issues/81 and https://github.com/swift-aws/aws-sdk-swift/issues/74


adam-fowler:
So far these changes have been tested extensively with SNS, SQS, S3 and a small subsection of EC2, SES and IAM.

PR #87, opened 2019-06-22
Avoid warning on Xcode

giginet:
Avoid warning on Xcode 11. This variable should be immutable.

Screen Shot 2019-06-22 at 21 43 50


giginet:
This CI result doesn't make sense... 😞

adam-fowler:
This looks fine

PR #89, opened 2019-07-02
XmlNode replacement

adam-fowler:
This is my alternative to the Foundation XMLNode classes. I implemented these for iOS as the Foundation classes are not available on iOS. 

Given the flakiness of the XML Foundation classes on Linux, we could use these in the main project as well. What are your thoughts?


Yasumoto:
Whoa, it's not on `iOS`? I see `XMLNode` looks like it's `macOS` only from the Foundation docs... 😮 

I looked at [`XMLParser`](https://developer.apple.com/documentation/foundation/xmlparser) but that looks like it's largely getting you the data as it moves through the document, which is quite unwieldy from my view.

Ideally we leverage Foundation for this (or make Foundation better) but can't think of a way to better support iOS 🤔 

What if we stuck with Foundation for this, and put this into the `apple` fork of the core? I'm not opposed to including this here, but would prefer not to maintain it if possible.

adam-fowler:
@Yasumoto Not sure what you mean about XMLParser, I use it to build the XMLNode tree, but that is the only place it is used.

Yasumoto:
Sorry about that, I meant to say something along the lines of "I see that we have access to `XMLParser` on iOS, macOS, and Linux, so I see why you used it. The tradeoff is we need to provide our own XML implementation (the `XML.swift` file)."

I think as much as we can use the same codepath across platforms the better, so if @adam-fowler you have a clear need for iOS then I'm in favor here. Thanks for the test coverage too!

adam-fowler:
iOS was something I said I was happy to support. If that means the XML code as well I am happy to do that. And yes it would be preferable that we have one solution across all the platforms.

But that isn't the only reason I was suggesting using our own XML library. The main reason was my confidence in the XML implementation in the Linux version of Swift is pretty low. I have had to implement workarounds in the aws-sdk-swift codebase to get around crash bugs in the library. Also the Linux version produces different output from the Darwin version, specifically in the xml document header.

Sources/AWSSDKSwiftCore/XML.swift
```diff
@@ -0,0 +1,465 @@
+//
+//  XML.swift
+//  AWSSDKSwift
+//
+//  Created by Adam Fowler on 2019/06/15
+//
+//
+import Foundation
+
+public class XML {
```
Yasumoto: Did you read any docs and/or RFCs to help you implement this? If so, linking here would be awesome!

Also, if you can add some of the context you shared (issues with the version in Foundation, missing iOS support) here as well, that would help set future-readers down the right path.

Sources/AWSSDKSwiftCore/XML.swift
```diff
@@ -0,0 +1,465 @@
...
+                    string += ":\(name)"
+                }
+                string += "=\"\(stringValue ?? "")\""
+                return string
+            default:
+                return ""
+            }
+        }
+        
+        /// CustomStringConvertible protocol
```
Yasumoto: You could move these into extensions if you'd like, that would help convey this meaning via code, which is nice.

```swift
extension XML: CustomStringConvertible {
```

Sources/AWSSDKSwiftCore/Message/Body.swift
```diff
@@ -55,53 +55,47 @@ extension Body {
...
-            } else if let element = node as? XMLElement {
-                let xmlDocument = XMLDocument(rootElement: element)
-                xmlDocument.version = "1.0"
-                xmlDocument.characterEncoding = "UTF-8"
-                return xmlDocument.xmlData
-            }
-            return nil
-            
+            let xmlDocument = XML.Document(rootElement: node)
+            xmlDocument.version = "1.0"
```
Yasumoto: Could/should those be set in the initializer if not passed through by default? 🤔

adam-fowler: My initial implementation was designed to stick to the Foundation implementation as tight as possible so moving between the two would be easy as possible. This is how it is implemented in the Foundation. I have already diverged slightly from the Foundation API, so can do so more.

adam-fowler: I removed these two lines as they aren't needed since the class output defaults to these values

Sources/AWSSDKSwiftCore/XML.swift
adam-fowler: Done

Sources/AWSSDKSwiftCore/XML.swift
adam-fowler: Done

Yasumoto:
🎊

Sources/AWSSDKSwiftCore/Message/Body.swift
Yasumoto: Awesome, thanks!

PR #91, opened 2019-07-05
Add missing Region: "eu-west-3"

alak:


adam-fowler:
Added https://github.com/swift-aws/aws-sdk-swift-core/commit/2afa2456d3bf4947e9e1322e408d2a129e5b4c59 instead which includes ap-east-1 and eu-north-1 as well.

alak:
Thanks @adam-fowler

PR #96, opened 2019-07-08
Deal with hostnames which include port in them

adam-fowler:
If port is included in the hostname that is given to the HttpClient then
ensure that is included in the "Host" entry of the headers. 

This fixes https://github.com/swift-aws/aws-sdk-swift-core/issues/94
I have verified this code works with standard AWS requests and also that it works with fakes3 and dynamodb_local.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -122,13 +122,7 @@ extension AWSClient {
     }
 
     private func createHTTPClient(for nioRequest: Request) -> HTTPClient {
-        let client: HTTPClient
-        if let _ = self._endpoint {
-            client = HTTPClient(hostname: nioRequest.head.host!, port: nioRequest.head.port ?? 443)
-        } else {
-            client = HTTPClient(hostname: nioRequest.head.hostWithPort!, port: 443)
-        }
-        return client
+        return HTTPClient(hostname: nioRequest.head.hostWithPort!, port: nioRequest.head.port ?? 443)
     }
 }
 
--- a/Sources/AWSSDKSwiftCore/HTTPClient.swift
+++ b/Sources/AWSSDKSwiftCore/HTTPClient.swift
@@ -95,6 +95,7 @@ private class HTTPClientResponseHandler: ChannelInboundHandler {
 
 public final class HTTPClient {
     private let hostname: String
+    private let headerHostname: String
     private let port: Int
     private let eventGroup: EventLoopGroup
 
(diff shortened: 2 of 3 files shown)


Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -122,13 +122,7 @@ extension AWSClient {
...
 
     private func createHTTPClient(for nioRequest: Request) -> HTTPClient {
-        let client: HTTPClient
-        if let _ = self._endpoint {
-            client = HTTPClient(hostname: nioRequest.head.host!, port: nioRequest.head.port ?? 443)
-        } else {
-            client = HTTPClient(hostname: nioRequest.head.hostWithPort!, port: 443)
-        }
-        return client
+        return HTTPClient(hostname: nioRequest.head.hostWithPort!, port: nioRequest.head.port ?? 443)
```
valeriomazzeo: Should the default port switch against the scheme? If http should be 80 rather than always assuming 443 for https.

Sources/AWSSDKSwiftCore/HTTPClient.swift
```diff
@@ -106,19 +107,26 @@ public final class HTTPClient {
...
+        
         self.hostname = hostname
-        self.port = port
+        
+        if let port = url.port {
+            self.port = port
+            self.headerHostname = "\(hostname):\(port)"
+        } else {
+            let isSecure = (scheme == "https")
+            self.port = isSecure == true ? 443 : 80
```
valeriomazzeo: self.port = isSecure ? 443 : 80

Sources/AWSSDKSwiftCore/AWSClient.swift
adam-fowler: As I understand it all aws access is https, so we can assume 443. Ideally we should be extracting all this info from a URL but I wanted to make as small a change as possible because this isn't my code.

Sources/AWSSDKSwiftCore/HTTPClient.swift
adam-fowler: fair enough

Sources/AWSSDKSwiftCore/AWSClient.swift
valeriomazzeo: I understand. It caught my attention because when I tried to workaround this issue, I tried to expose the fake SQS service to port `80` so that `Host: localhost` would match (before this PR), but I was unable to make it work because of this assumption.

Maybe it can be discussed in another pull request?

adam-fowler: It's certainly an area that could be cleaned

jonnymacs:
@adam-fowler just making sure port 80 for http and port 443 for https should not be included in the header. Only if the port is something other than the default.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

I think the way I read this code the host will now always include the port in the header even if it is the default port, which is not right. (haven't tested, so forgive me if I am misreading)

Given we haven't had any issues with this until now I'd prefer not to change this behavior for the default ports. Specifically the IAM credential services are finicky with these values.

adam-fowler:
@jonnymacs no it only includes the port in the header if the port is explicitly included in the host name eg localhost:9324. When working with AWS and not a fake service no port is passed in so the port is not included in the header. Anyway it would fail if the port was included as the signing of the request would be wrong.

Given @noppoMan recently fixed an issue with this code I don't think this has always worked though.

jonnymacs:
@adam-fowler if you set hostname to https://something:443 in the HTTPClient hostname / port initializer then the :443 will be included in the header.

I understand this is not happening, but the HTTPClient init api is not quite right in this case.

Let me know if that makes sense?

What I am asking is for the HTTPClient to ensure the host header does not include 80 or 443 for http / https even if the hostname includes it on init.

adam-fowler:
@jonnymacs I can add that if you'd prefer, It seems a bit of an edge case. Also it would break the signing code which includes the port if it is specified.

jonnymacs:
@adam-fowler yeah I think i'd prefer.  Just thinking about the client as an independent module that can be used for more than just signed requests (like calls to the IAM meta data service)

adam-fowler:
ok will do, this will require changes to the signing code as well

PR #99, opened 2019-07-21
Add x-amz-security-token header before calling authorization()

adam-fowler:
The "x-amz-security-token" header value was being added after the "Authorization" header value was calculated. This was then giving the error
```
accessDenied(message: Optional("RequestId: 4059FAA108A45F89, 
    HeadersNotSigned: x-amz-security-token, 
    Message: There were headers present in the request which were not signed, 
    HostId: p013uVEP7DHgqGkbCe/zsjuASVE0b9h9a7Gb37OfnuPmFnNeMZ5VnILzvi9orDLzHtfn1yNEzLA="))
```

This pull request moves the adding of the "x-amz-security-token" header value to before the authorization value calculation. This fixes https://github.com/swift-aws/aws-sdk-swift/issues/151

### DIFF
--- a/Sources/AWSSDKSwiftCore/Signers/V4.swift
+++ b/Sources/AWSSDKSwiftCore/Signers/V4.swift
@@ -128,6 +128,10 @@ extension Signers {
                 headersForSign[header.key] = header.value
             }
 
+            if let token = credentialForSignature.sessionToken {
+                headersForSign["x-amz-security-token"] = token
+            }
+            
             headersForSign["Authorization"] = authorization(
                 url: url,
                 headers: headersForSign,
@@ -136,10 +140,6 @@ extension Signers {
                 bodyDigest: bodyDigest,
                 credentialForSignature: credentialForSignature
             )
-
-            if let token = credentialForSignature.sessionToken {
-                headersForSign["x-amz-security-token"] = token
-            }
 
             return headersForSign
         }


jonnymacs:
@adam-fowler I am very confused. I've been using ECS credentials for over a year with Kinesis which include a sessionToken, and don't get an error...

jonnymacs:
Docs are not _clear_ but, seems like it should be signed https://docs.amazonaws.cn/en_us/general/latest/gr/sigv4-create-canonical-request.html

You can use temporary security credentials provided by the AWS Security Token Service (AWS STS) to sign a request. The process is the same as using long-term credentials, but when you include signing information in the Authorization header you must add an additional HTTP header for the security token. The header name is X-Amz-Security-Token, and the header's value is the session token (the string you received from AWS STS when you obtained temporary security credentials).

jonnymacs:
I wonder if the source of the token matters whether it is signed...

adam-fowler:
@jonnymacs Could you test your ECS credentials work with the change? Could it be related to the service being used. In the original bug it was indicated it worked for SNS, SQS but not S3. Maybe Kinesis doesn't need it signed. I've replicated the issue in S3 and also seen that SNS doesn't need the change.

jonnymacs:
@adam-fowler yep. i'm going to test today. stay tuned

adam-fowler:
Hey @jonnymacs maybe you could try to call S3.PutObject() in your ec2 instance with the current setup. See if that works

jonnymacs:
@adam-fowler i'm not sure I'll have time to regress putObject with the latest tag. I am comfortable moving fwd with this since it seems like it's right and should have been like this all along

PR #101, opened 2019-07-21
Add AWSClient.init() which takes a Credential parameter

adam-fowler:
STS.AssumeRole() returns a credential structure, which contains a sessionToken. There is currently no way to initialize an AWS service with this sessionToken. 

This PR is the starting point of changing that. I have added a new AWSRequest.init() which takes a Credential structure as a parameter. The original initializer now creates a Credential structure and passes it onto the new initializer. Once this is in I can change the CodeGenerator to output code to create services taking a Credential structure.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -68,10 +68,10 @@ public struct AWSClient {
 
     public static let eventGroup: EventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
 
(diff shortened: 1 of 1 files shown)


jonnymacs:
@adam-fowler I think you mean AWSClient.init right? (not AWSRequest...)

I think I agree with the spirit of this.

Curious, why not just support an optional token in the current client initializer? Then pass it to the Credential initializer, where it's already optional.

```
if let accessKey = accessKeyId, let secretKey = secretAccessKey {
        credential = Credential(accessKeyId: accessKey, secretAccessKey: secretKey, sessionToken: sessionToken)
} else if...
```

adam-fowler:
Yes I meant AWSClient.
Yeah I guess I originally thought I might need to pass on the full Credential structure. I'll change to just passing in an optional token

adam-fowler:
@jonnymacs do you prefer the new version of the code?

PR #103, opened 2019-08-01
DictionaryDecoder: change how we deal with expecting an array

adam-fowler:
Previously if we were expecting an array and didn't get an array we would throw an error. This change will place the object we found inside an array and process that.

This fixes an issue in apigateway https://github.com/swift-aws/aws-sdk-swift/issues/33 where single values can be returned even though we are expecting an array. Not totally sure about this change, as it is slightly weird. Although it shouldn't break anything as it is just giving the decoder another chance to decode the response, when it fails because it didn't find an array.


adam-fowler:
Yeah its an annoyance. If you have two values api gateway will return an array, but when there is only one it doesn't

PR #105, opened 2019-08-08
OpenSSL 1.1

adam-fowler:
Get aws-sdk-swift working with OpenSSL 1.1 without breaking earlier versions of OpenSSL.

Added CAWSSDKOpenSSL target to shim over differences in OpenSSL versions, specifically how HMAC contexts are created.

This is only needed for versions of the library that use nio 1.4 or less as nio2.0 includes BoringSSL with it.

@Yasumoto this is from the conversation on discord #aws this morning/yesterday


jonnymacs:
tested on OS X and Xenial (ubuntu 16.04)

LG 💯

PR #108, opened 2019-08-21
Check a match can be found in pattern validation

adam-fowler:
Check a match can be found in pattern validation, not that the whole string matches the pattern.

It appears some services expect this eg MediaConvert. Unfortunately this makes the validation slightly less rigourous for other services. We still expect the pattern to be anchored to the start of the string, so maybe this won't be too much of a change.


adam-fowler:
of course now I find this is causing issues else where. I noticed aws-sdk-go don't do the pattern matching, maybe they gave up.

PR #109, opened 2019-08-21
Logging middleware

adam-fowler:
An AWSServiceMiddleware that outputs the contents of requests and responses as they are sent and received from AWS.


adam-fowler:
ah messed up the branch, restarting

PR #111, opened 2019-08-25
Logging middleware

adam-fowler:
An AWSServiceMiddleware that outputs the contents of requests and responses as they are sent and received from AWS.

Along with https://github.com/swift-aws/aws-sdk-swift/pull/162/files you can create a service as follows and have it log all requests to it and the corresponding responses.

```let s3 = S3(region:.useast1, middlewares:[AWSLoggingMiddleware()])```

### CODE REVIEW

File: Sources/AWSSDKSwiftCore/Message/AWSMiddleware.swift [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+public extension AWSServiceMiddleware {
+    func chain(request: AWSRequest) throws -> AWSRequest {
+        return request
+    }
+    func chain(responseBody: Body) throws -> Body {
+        return responseBody
+    }
+}
+
+#if DEBUG
```
jonnymacs: In our shared dev and qa environments, we run code in production mode. It would be cool if I could enable this somehow in those envs if I wanted to. 

Maybe this code should compile in non debug envs, then allow setting the log level in the init

```
    let logLevel : LogLevel

    init(log : @escaping (String)->() = { print($0) }, logLevel: LogLevel = .info) {
        self.log = log
        self.logLevel = logLevel
    }
    
    enum LogLevel {
      case debug
      case info
      case error
    }

    /// output request
    public func chain(request: AWSRequest) throws -> AWSRequest {
        if logLevel == .debug {
          log("Request: \(request.operation)")
          log("\(request.httpMethod) \(request.url)")
          log("Headers: " + getHeadersOutput(request.httpHeaders))
          log("Body: " + getBodyOutput(request.body))
        }
        return request
    }

    /// output response
    public func chain(responseBody: Body) throws -> Body {
        if logLevel == .debug {
            var output = "Response: "
            output += getBodyOutput(responseBody)
            log(output)
        }
        return responseBody
    }

----


    let kinesisClient = AWSClient(
        accessKeyId: "foo",
        secretAccessKey: "bar",
        region: nil,
        amzTarget: "Kinesis_20131202",
        service: "kinesis",
        serviceProtocol: ServiceProtocol(type: .json, version: ServiceProtocol.Version(major: 1, minor: 1)),
        apiVersion: "2013-12-02",
        middlewares: [AWSLoggingMiddleware(logLevel: .debug)],
        possibleErrorTypes: [KinesisErrorType.self]
    )
```

I can update this PR if you like this idea

adam-fowler: code can be made available in release. What would you output for logLevel info and error though? If there is no difference then the way to disable the logging would be not include the middleware.

jonnymacs:
Info and error would be for future use.

I’d like to be able to include the middleware, but dynamically set the log level based on my own business logic. In this way I can programmatically enable / disable.

I suppose one could programmatically init and include the middleware or not. Just seems cleaner to set the log level on the middleware... not a huge deal like I said, tol (take or leave)

adam-fowler:
ok fair enough. I'll merge this pull request and you can add in a new PR with your changes

PR #116, opened 2019-09-02
Add public facing signURL() function to AWSClient

adam-fowler:
Clean up code path to generating signed URLs. See https://github.com/swift-aws/aws-sdk-swift/issues/73

```
let s3 = S3(.euwest1)
let url = "https://mybucket.s3.amazonaws.com/key/for/my/file"
let signedURL = s3.client.signURL(url: url, httpMethod: "GET")
```


adam-fowler:
If you're talking about the removal of public from signer, that is reverting the code to what it was at the point of the previous release. Making signer public was another way to give people access to signing urls, which in the end I decided against

PR #120, opened 2019-09-12
Documentation for aws-sdk-swift-core

adam-fowler:
- Scripts to build documentation for aws-sdk-swift-core
- Cleaned up code, where documentation indicated some mess.
- Removed public from some objects that don't need to be public
- Removed unused functionality
You can see the documention here https://adam-fowler.github.io/aws-sdk-swift-core/docs/current/index.html


jonnymacs:
merged latest master and tested with kinesis, lgtm!

PR #121, opened 2019-09-13
AWSClient.send() functions don't throw errors so don't mark them as if they do

adam-fowler:
This will simplify using the library a little

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -118,7 +118,7 @@ extension AWSClient {
 
 // public facing apis
 extension AWSClient {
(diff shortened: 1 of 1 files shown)


Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -151,7 +151,7 @@ extension AWSClient {
             }
     }
 
-    public func send<Output: AWSShape>(operation operationName: String, path: String, httpMethod: String) throws -> Future<Output> {
+    public func send<Output: AWSShape>(operation operationName: String, path: String, httpMethod: String) -> Future<Output> {
 
         return signer.manageCredential().thenThrowing { _ in
```
Yasumoto: Great save! `thenThrowing`, renamed to [`flatMapThrowing`](https://github.com/apple/swift-nio/blob/master/Sources/NIO/EventLoopFuture.swift#L520) in NIO2, does not throw, but returns a failed future on error 🎉

PR #127, opened 2019-10-15
HTTP Client tidy and iOS/NIOTransportServices

adam-fowler:
This ones been waiting until nio2.0 went in. Sorry combined a number of changes in one PR. Before you say anything about async-http-client I looked at just replacing our client with it but found that it would kill iOS builds so wasn't really that keen. Also macOS builds would be forced to use the Sockets version of NIO instead of Network.framework version which is preferable. Anyway changes include

- Moved the HTTP request setup into a HTTPClientRequestSerializer class and out of ClientBootstrap code. Similar to how other http clients are doing. Its looks much cleaner.
- HTTPClient.init() no longer needs the host details. Although we don't do this could allow us to re-use the same HTTPClient for multiple calls.
- Implemented EventLoopGroupProvider enum and initialise HTTPClient with one.
- Support NIOTransportServices versions of EventLoopGroup and ClientBootstrap. These are only available to AppleOS builds. Code to recognise this is #if canImport(Network) as recommended by NIO team. Also need to check correct versions of OS are available. Doing this mean aws-sdk-swift now supports macOS, Linux and iOS. We can archive the aws-sdk-appleos repositories as they are no longer needed.


Sources/AWSSDKSwiftCore/HTTPClient.swift
```diff
@@ -4,190 +4,276 @@
...
+            }
+        #else
+            clientBootstrap(hostname: hostname, port: port, headerHostname: headerHostname, request: request, response: response)
+        #endif
+        
         return response.futureResult
     }
 
+   /// shutdown the event group
     public func close(_ callback: @escaping (Error?) -> Void) {
```
Yasumoto: Maybe we should also [create a `deinit` where we check to make sure that `close` was called](https://github.com/swift-server/async-http-client/blob/master/Sources/AsyncHTTPClient/HTTPClient.swift#L68) ?

adam-fowler:
going to split this request into two

PR #130, opened 2019-10-21
HTTPClient changes

adam-fowler:
- Moved the HTTP request setup into a HTTPClientRequestSerializer class and out of ClientBootstrap code. Similar to how other http clients are doing. Its looks much cleaner.
- HTTPClient.init() no longer needs the host details. Although we don't do this, could allow us to re-use the same HTTPClient for multiple calls.
- Implemented EventLoopGroupProvider enum and initialise HTTPClient with one.
- Use AWSClient.eventGroup in HTTPClient

### DIFF
--- a/Sources/AWSSDKSwiftCore/MetaDataService.swift
+++ b/Sources/AWSSDKSwiftCore/MetaDataService.swift
@@ -55,8 +55,8 @@ struct MetaDataService {
 
         func uri() throws -> Future<String> {
             switch self {
-            case .ecsCredentials(let containerCredentialsUri):
-                return AWSClient.eventGroup.next().makeSucceededFuture(containerCredentialsUri)
+            case .ecsCredentials(_):
+                return AWSClient.eventGroup.next().makeSucceededFuture(self.baseURLString)
             case .instanceProfileCredentials:
                 // instance service expects absoluteString as uri...
                 return MetaDataService.request(host: host, uri: baseURLString, timeout: 2).flatMapThrowing{ response in
--- a/Sources/AWSSDKSwiftCore/MetaDataService.swift
+++ b/Sources/AWSSDKSwiftCore/MetaDataService.swift
@@ -89,7 +89,7 @@ struct MetaDataService {
     }
 
     static func request(host: String, uri: String, timeout: TimeInterval) -> Future<HTTPClient.Response> {
-        let client = HTTPClient(hostname: host, port: 80)
+        let client = HTTPClient()
         let head = HTTPRequestHead(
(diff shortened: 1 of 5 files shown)


Sources/AWSSDKSwiftCore/MetaDataService.swift
```diff
@@ -89,7 +89,7 @@ struct MetaDataService {
     }
 
     static func request(host: String, uri: String, timeout: TimeInterval) -> Future<HTTPClient.Response> {
-        let client = HTTPClient(hostname: host, port: 80)
```
jonnymacs: If we are relying on the uri for the host / port info, then I think MetaDataServiceProvider#uri needs to return the baseURLString for .ecsCredentials instead of the containerCredentialsUri which does not include the host / port.

adam-fowler: The HTTPClient will default to port 80 if an https:// scheme prefix isn't found so this should work ok. Haven't setup an ECS container before so not sure what we get back from "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI" environment variable. Does it not include the host info?

jonnymacs: correct `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is just the path.

adam-fowler: Does the change fix this? 

The MetaDataServiceProvider enum is a bit confusing. baseURLString, uri mean different things for different providers. Would be better as a protocol and individual classes
```
protocol MetaDataServiceProvider {
    public static func getCredential() throws -> Future<CredentialProvider>
}

extension MetaDataServiceProvider {
    static func request(host: String, uri: String, timeout: TimeInterval) -> Future<HTTPClient.Response> {
        ...
    }
}

class ECSMetaDataService: MetaDataServiceProvider {
    public static func getCredential() throws -> Future<CredentialProvider> {
    }
}

class InstanceProfileMetaDataService: MetaDataServiceProvider {
    public static func getCredential() throws -> Future<CredentialProvider> {
    }
}

```

jonnymacs: @adam-fowler yes, the change fixes it.

Agreed it's confusing. I like your suggestions. Feel free, and I'll put it on my todo as well.

adam-fowler: @jonnymacs You good for this one to go in?

PR #131, opened 2019-10-21
NIO Transport Services

adam-fowler:
- Support NIOTransportServices versions of EventLoopGroup and ClientBootstrap. These are only available to platforms that have the Network.framework ie macOS and iOS. Code to recognise this is #if canImport(Network) as recommended by NIO team. Also need to check correct versions of OS are available. Doing this mean aws-sdk-swift now supports macOS, Linux and iOS. We can archive the aws-sdk-appleos repositories as they are no longer needed.

### DIFF
--- a/Sources/AWSSDKSwiftCore/HTTPClient.swift
+++ b/Sources/AWSSDKSwiftCore/HTTPClient.swift
@@ -36,12 +36,25 @@ public final class HTTPClient {
         case malformedHead
         case malformedBody
         case malformedURL(url: String)
+        case alreadyShutdown
     }
 
-    private let eventLoopGroup: EventLoopGroup
+    /// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
+    public enum EventLoopGroupProvider {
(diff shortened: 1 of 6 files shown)


Sources/AWSSDKSwiftCore/HTTPClient.swift
```diff
@@ -4,190 +4,294 @@
...
+        return pipeline.eventLoop.makeSucceededFuture(())
+    }
 
-        _ = ClientBootstrap(group: eventGroup)
+    /// setup client bootstrap for HTTP request
+    func clientBootstrap(hostname: String, port: Int, headerHostname: String, request: Request, response: EventLoopPromise<Response>) {
+        _ = ClientBootstrap(group: self.eventLoopGroup)
             .connectTimeout(TimeAmount.seconds(5))
-            .channelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+            .channelOption(ChannelOptions.socket(SocketOptionLevel(IPPROTO_TCP), TCP_NODELAY), value: 1)
```
Yasumoto: 👍

PR #133, opened 2019-10-22
Disable tests that don't compile in release, which compiling release

adam-fowler:


Yasumoto:
I like https://github.com/swift-aws/aws-sdk-swift-core/pull/134 more, since that keeps test-coverage the same. However, I'm open-minded!

adam-fowler:
Deleted in favour of #134

Issue #43, opened 2018-12-06
tests don't run in release mode


weissi:
I got this compilation failure when running the tests with

    swift test --sanitize=address -c release -Xswiftc -enable-testing

which we do as part of the SwiftNIO compatibility/test suite before releases.

```
Compile Swift Module 'AWSSDKSwiftCore' (29 sources)
Compile Swift Module 'AWSSDKSwiftCoreTests' (9 sources)
/private/tmp/.nio-release-tools_z0dmMg/aws-sdk-swift-core/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift:39:34: error: value of type 'AWSClient' has no member 'debugCreateAWSRequest'
            let awsRequest = try sesClient.debugCreateAWSRequest(
                                 ^~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~
/private/tmp/.nio-release-tools_z0dmMg/aws-sdk-swift-core/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift:66:34: error: value of type 'AWSClient' has no member 'debugCreateAWSRequest'
            let awsRequest = try kinesisClient.debugCreateAWSRequest(
                                 ^~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~
error: terminated(1): /Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-build-tool -f /private/tmp/.nio-release-tools_z0dmMg/aws-sdk-swift-core/.build/release.yaml test output:
```

jonnymacs:
@weissi hmm... yeah this is to maintain the fileprivate status of createAWSRequest method in release mode. The public debug method is added as an extension when the Debug directive is set

`-Xswiftc -DDEBUG`

is this an option for you to add?

weissi:
@jonnymacs sure, that's fine. Just out of interest, why don't you make it `internal` ?

jonnymacs:
@weissi it's a good question.  I didn't define the original access control for createAWSRequest, that was @noppoMan, so he may provide a better answer. But I imagine it's because internal still allows for exposing the method through an extension in another part of the code and we want to restrict all access to this method to just this file.

Using the DEBUG directive to add an internal debug method that is exposed with @testable seemed like a nice workaround to preserve the fileprivate access control while still adding some tests. (note - i misspoke in my prev comment, debugCreateAWSRequest is not public)

[...]

PR #134, opened 2019-10-23
Remove the debugValidate and debugCreateAWSRequest

Yasumoto:
We still control the usage of these methods when protected with internal, and this removes extra code during debug vs. release builds

Let's fixup https://github.com/swift-aws/aws-sdk-swift-core/issues/43

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -312,7 +312,7 @@ extension AWSClient {
         }
     }
 
-    fileprivate func createAWSRequest(operation operationName: String, path: String, httpMethod: String) throws -> AWSRequest {
+    internal func createAWSRequest(operation operationName: String, path: String, httpMethod: String) throws -> AWSRequest {
 
         guard let url = URL(string: "\(endpoint)\(path)"), let _ = url.hostWithPort else {
             throw RequestError.invalidURL("\(endpoint)\(path) must specify url host and scheme")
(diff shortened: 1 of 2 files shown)


jonnymacs:
@Yasumoto we lose some control with internal, because now I can call these methods from an extension if I want to create one for the AWSClient. Maybe this is not a big deal, but fileprivate protection is a feature of swift and we are using it. So I'm not a fan of this change.  Testing fileprivate methods is a thing and the debug directive supports testing them.

Not testing fileprivate methods, or insisting tests all run in release mode is a real head scratcher to me. I want to be supportive for #43, but I'm not sure why this is a requirement.

I've proposed to @adam-fowler that we can wrap certain tests in the debug directive as well. This will resolve #43. I think we do a good job maintaining our code and we test our code in test mode so the tests will still get run.

Yasumoto:
Totally feel ya, Adam has https://github.com/swift-aws/aws-sdk-swift-core/pull/133 out for review to solve this in the way y'all suggested!

I think both ways are valid, I lean toward "I'd prefer tests to run in both debug & release" but totally feel ya that there are tradeoffs. If it's 2 v. 1 in this case I'm happy to close this and go the other route 👍

adam-fowler:
I'm not too bothered either way but one point to make, if you create an extension of AWSClient outside of aws-sdk-swift-core it only has access to the public elements of AWSClient.

jonnymacs:
@adam-fowler you're right. 

`@testable` makes private functions accessible in an extension. This is what threw me off.

OK. So fileprivate seems like it's the wrong protection level here then in the first place. The point of fileprivate is to make methods accessible from another class but only when accessed from within the same file.

I stand by using the debug directive for testing fileprivate methods should be tolerated, but considering we don't need fileprivate in this case, I think this PR is the way to go.

Thnx for bearing with.

Yasumoto:
Thanks for working through this, folks! (And thanks for merging since I got tied up with work 😅 )

PR #136, opened 2019-10-30
Provide a path for initialising a service without credentials

adam-fowler:
CognitoIdentity and CognitoIdentityProvider can be used without authentication details. These changes allow for this to happen. 

- In AWSClient.init() If accessKeyId and secretAccessKey are nil create an expired credential.
- Metadata service is only used when credentials have expired
- Don't sign requests with empty credentials
- This means you can initialize AWSClient with accessKeyId: "", secretAccessKey: "" and avoid asking for EC2/ECS credentials and not sign requests
- CognitoIdentityProvider works without signed credentials
- Removed invalid signing test

### DIFF
--- a/Sources/AWSSDKSwiftCore/Signers/V4.swift
+++ b/Sources/AWSSDKSwiftCore/Signers/V4.swift
@@ -49,11 +49,7 @@ extension Signers {
         public func manageCredential() -> Future<CredentialProvider> {
             if credential.nearExpiration() {
                 do {
-                    return try MetaDataService.getCredential()
-                        .flatMapError { _ in
-                            return AWSClient.eventGroup.next().makeSucceededFuture(Credential(accessKeyId: "", secretAccessKey: ""))
-                        }
-                        .map { credential in
+                    return try MetaDataService.getCredential().map { credential in
                             self.credential = credential
                             return credential
                     }
--- a/Sources/AWSSDKSwiftCore/Signers/V4.swift
+++ b/Sources/AWSSDKSwiftCore/Signers/V4.swift
@@ -47,6 +47,7 @@ extension Signers {
 
         /// If you did not provide credentials `manageCredential()` should be called and the future resolved prior to building signedURL or signedHeaders to ensure latest credentials are retreived and set
         public func manageCredential() -> Future<CredentialProvider> {
+#if os(Linux)
             if credential.nearExpiration() {
                 do {
                     return try MetaDataService.getCredential().map { credential in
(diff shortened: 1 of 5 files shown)


Sources/AWSSDKSwiftCore/Signers/V4.swift
```diff
@@ -47,11 +47,15 @@ extension Signers {
...
         /// If you did not provide credentials `manageCredential()` should be called and the future resolved prior to building signedURL or signedHeaders to ensure latest credentials are retreived and set
         public func manageCredential() -> Future<CredentialProvider> {
-            if credential.isEmpty() || credential.nearExpiration() {
+            if credential.nearExpiration() {
                 do {
-                    return try MetaDataService.getCredential().map { credential in
-                        self.credential = credential
-                        return credential
+                    return try MetaDataService.getCredential()
+                        .flatMapError { _ in
```
jonnymacs: Just trying to figure out this addition. I think this means not setting access key or secret at all will lead to skip signing if the metadata service fails. Is this what we want? Im not sure it is. I’d rather surface the error so the client knows it tried to get creds from the metadata service. But if it is why not set the credential instance variable on this error handler as well so nearExpiration is always false moving fwd. then we wont try the metadata service again the next time?

adam-fowler: The reason for this change is there are some aws services that don’t require credentials (eg CognitoIdentityProvider) but we are requiring credentials all the time. This change is to stop the metadata service request if you explicitly set the credentials to empty. 

I’m actually wondering if we should disable the metadata service calls completely on non Linux builds because they don’t make sense outside of ECS/EC2 instances.

adam-fowler:
On the Linux if the user doesn’t provide credentials they are given an expired credentials which will the client will try to update through the metadata service. If it fails it should throw an error as it has previously. Don’t have code directly in front of me so can’t be sure but that was what I hoped.

jonnymacs:
Ok yeah. I understand the overall purpose of this pr, just asking about the addition of flatMapError returns a succeeded future with empty credentials when the call to metadata service fails.

adam-fowler:
Ok let me review when I have access to the code. 
What’s your thoughts on disabling the metadata service for non Linux platforms?

adam-fowler:
Yeah sorry, was code I had put in initially when trying to work out best way to differentiate between requests that require the metadate service and requests that don't want credentials. It has been removed.

adam-fowler:
@jonnymacs you good for this to go in now? With the metadata service code removed for non-Linux builds?

Issue #137, opened 2019-11-04
Option to inject `EventLoopGroup` into `AWSClient`


fabianfett:
**Describe the bug**
The `AWSClient` currently holds a static shared `EventLoopGroup` instance. 

```swift
public static let eventGroup: EventLoopGroup = createEventLoopGroup()
```

**Expected behavior**

As Cory Benfield [pointed out](https://forums.swift.org/t/how-to-wrap-the-initialization-ceremony-with-a-client/14541/2):

> Libraries should almost never choose to own their own EventLoopGroup. This is because it's highly valuable to have an extremely small number of groups in an application; in fact, for many applications you may want only one.

Therefore I would assume, one can have the chance to inject ones own `EventLoopGroup` into the `AWSClient`. Therefore I propose we add another property `eventLoopGroup` to the quite long initializer:

```swift
public init(accessKeyId: String? = nil, secretAccessKey: String? = nil, sessionToken: String? = nil, region givenRegion: Region?, amzTarget: String? = nil, service: String, signingName: String? = nil, serviceProtocol: ServiceProtocol, apiVersion: String, endpoint: String? = nil, serviceEndpoints: [String: String] = [:], partitionEndpoint: String? = nil, middlewares: [AWSServiceMiddleware] = [], possibleErrorTypes: [AWSErrorType.Type]? = nil)
```

**Additional context**

When thinking about solving this problem there are two possible implementation options:

1. The `eventLoopGroup` has a default value at initialization (the current `AWSClient.eventGroup`)
- **Pro**: Non breaking API Change
- **Con**: The style does not enforce minimizing the number of `EventLoopGroup`s.
2. The `eventLoopGroup` must be provided at initialization
- **Pro**: The style does enforce minimizing the number of `EventLoopGroup`s. 
- **Pro**: The shared instance `AWSClient.eventGroup` could be removed.
- **Con**: A breaking API Change

**Personal opinion**

Since a 4.0.0 release is around the corner, we could go for the breaking API change. I think that a library as important as `aws-sdk-swift` should embrace using as few `EventLoopGroup`s as possible. But I could understand, if such a big API breaking change is not intended for now.

However the maintainers decide, I would be more than happy to provide a PR for this.

fabianfett:
Also this change would make it much easier to use the `AsyncHTTPClient` in the future if wanted.

adam-fowler:
[...]

PR #139, opened 2019-11-05
Added an option to inject the EventLoopGroup into the AWSClient.

fabianfett:
PR for issue #137

### Notes:

In order to check if anything else was using `AWSClient.eventGroup` I deprecated `eventGroup`:

```swift
  @available(*, deprecated, message: "Consumers of this API shoud provide their own runtime at the initialization of an AWSClient with the EventLoopGroupProvider.use(eventLoopGroup) option.")
  public static let eventGroup: EventLoopGroup = _eventGroup
  private static let _eventGroup: EventLoopGroup = createEventLoopGroup()
```

This made me aware that `AWSClient.eventGroup` was also used in `Signers.V4` and `MetaDataService`. Therefore I had to dig deeper to change those APIs. I did so by dependency injection from `AWSClient`. Sadly those are breaking API changes (at least for Signers.V4 since this is a public class). It is to be discussed how many consumers use `Signers.V4` directly and how big of an impact this change is.

One could argue that we should avoid API Changes no matter what. On the other hand changing those APIs would allow us to reduce the impact of the shared `EventLoopGroup`.

I've also prepared the according changes for the `aws-sdk-swift` package. I will create a PR for this, as soon as this PR is accepted.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -147,7 +147,7 @@ public class AWSClient {
 // invoker
 extension AWSClient {
     fileprivate func invoke(_ nioRequest: HTTPClient.Request) -> Future<HTTPClient.Response> {
-        let client = HTTPClient(eventLoopGroupProvider: .shared(self.eventLoopGroup))
+        let client = HTTPClient(eventLoopGroupProvider: .shared(eventLoopGroup))
         let futureResponse = client.connect(nioRequest)
 
         futureResponse.whenComplete { _ in
@@ -175,7 +175,7 @@ extension AWSClient {
     ///     Empty Future that completes when response is received
     public func send<Input: AWSShape>(operation operationName: String, path: String, httpMethod: String, input: Input) -> Future<Void> {
 
-        return signer.manageCredential(eventLoopGroup: self.eventLoopGroup).flatMapThrowing { _ in
+        return signer.manageCredential(eventLoopGroup: eventLoopGroup).flatMapThrowing { _ in
                 let awsRequest = try self.createAWSRequest(
                     operation: operationName,
                     path: path,
@@ -199,7 +199,7 @@ extension AWSClient {
     ///     Empty Future that completes when response is received
     public func send(operation operationName: String, path: String, httpMethod: String) -> Future<Void> {
 
-        return signer.manageCredential(eventLoopGroup: self.eventLoopGroup).flatMapThrowing { _ in
+        return signer.manageCredential(eventLoopGroup: eventLoopGroup).flatMapThrowing { _ in
                 let awsRequest = try self.createAWSRequest(
                     operation: operationName,
                     path: path,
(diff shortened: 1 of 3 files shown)


adam-fowler:
I've just submitted a PR refactoring the metadata service, if you could base your changes off that, it be great. That change also deals a lot with the passing of eventLoopGroups around the various classes so your changes should only need need to be AWSClient specific.

fabianfett:
@adam-fowler I rebased and made the changes less intrusive. I'm looking forward to your feedback.

adam-fowler:
You missed the HTTPClient change I asked for. With that change on mac the user can control whether he uses NIOTransportServices or NIOSSL. Otherwise the user is forced into using NIOTransportServices and the code would crash if a vanilla EventLoopGroup was passed in.

Otherwise looks a lot cleaner.

fabianfett:
@adam-fowler sorry for missing that. It's now included.

adam-fowler:
@jonnymacs we should use the non static AWSClient.eventLoopGroup in the s3 multipart upload/download functions to ensure it is the same eventLoopGroup used in the client and of course we'll be able to remove the `hop(to:)` calls.

@jonnymacs @Yasumoto do we want the `eventLoopGroupProvider` parameter to have a default value? Given this is included in a new major release I'm thinking we can remove the default value. Regardless of how we feel about this we can remove the default value from AWSClient as this is initialised internally in the AWSSDK service classes. We can make that call at service class level.

fabianfett:
> @fabianfett thanks for joining the effort on this project. You're making some great observations so far, and this was one of them. We've been advising clients to hop(to:) their own eventLoop (if they are on one) when receiving futures from this aws client, but offering injection has always been the goal and now we'll have it.

Thanks @jonnymacs for the warm welcome.

> we should use the non static AWSClient.eventLoopGroup in the s3 multipart upload/download functions to ensure it is the same eventLoopGroup used in the client and of course we'll be able to remove the hop(to:) calls.

I already have a PR prepared for this. If you don't mind I would put it on the `aws-sdk-swift` repo next.

> Do we want the eventLoopGroupProvider parameter to have a default value? Given this is included in a new major release I'm thinking we can remove the default value.

I am very much in favor of this since it pushes the consumer into the direction of thinking about eventLoopGroups, which should ultimately make `swift-nio` code more performant, which, I guess, is our shared goal.

I have one more question about

`public static let sharedEventLoopGroup: EventLoopGroup = createEventLoopGroup()`

Do we really want to advertise the sharedEventLoopGroup publicly? What do you think about making this private or at least internal?

fabianfett:
> Then @fabianfett just one nit I have, a style thing - I don't personally feel the need to reference instance variables with self. on gets. In some cases it's necessary when in a closure or to disambiguate with a method argument, but in this PR I don't think you have to when passing eventLoopGroup as an argument to manageCredential or when initializing the HTTPClient. I've heard the it's more explicit argument, but I still don't prefer it.

This is an area for intense discussion that I don't want to get into. I want to do whatever brings this effort forward so I removed the `self.`. If I make any other commit on this repo, please don't take it personally if another `self.` surfaces. This is just how I'm used to write my code. But I don't have any problem with removing those.

adam-fowler:
I think we are good, merging going ahead
Welcome to aws-sdk-swift @fabianfett

adam-fowler:
@fabianfett I don't see any reason for sharedEventLoopGroup to be public. It was one of the things I noticed but forgot to mention. 

Do you want to stick in another PR for that, also you could remove the default value for eventLoopGroupProvider. As I said earlier we can make the call on whether that needs a default value at the service level.

adam-fowler:
@fabianfett have you got that aws-sdk-swift PR ready. Would be good to get in as currently aws-sdk-swift is not compiling with latest from core.

fabianfett:
> @fabianfett have you got that aws-sdk-swift PR ready. Would be good to get in as currently aws-sdk-swift is not compiling with latest from core.

@adam-fowler please review #144. Next up PR on aws-sdk-swift.

Yasumoto:
This is great! Thanks so much folks 🎉

PR #141, opened 2019-11-07
AsyncHTTPClient

adam-fowler:
So here it is. I eventually decided to take on integrating the SSWG [AsyncHTTPClient](http://github.com/swift-server/async-http-client) into aws-sdk-swift-core. This isn't final and will need a lot of testing but I thought I would put it up so everyone can have a look. I warn you there are a lot of changes. I had to keep our own HTTPClient around as well for the builds using NIOTransportServices ie iOS and currently macOS (although macOS can switch between the two). A lot of the work was setting up choosing between clients. SSWG are looking into supporting NIOTransportServices with AsyncHTTPClient https://github.com/swift-server/async-http-client/issues/127 so maybe eventually we can remove the selection code and our own HTTPClient altogether.

Below is a list of everthing I did.
(Edited to reflect changes discussed below)
- Created a protocol `AWSHTTPClient` defining a common interface for a [HTTP Client](https://github.com/swift-aws/aws-sdk-swift-core/blob/async-http-client/Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift). Set this up to hide implementation details of HTTP client from AWSClient. 
- Extended both our HTTPClient and the AsyncHTTPClient to conform with the AWSHTTPClient protocol.
- Renamed HTTPClient to `NIOTSHTTPClient` and removed all non NIO Transport services code from it. We are only using this client when we are using NIO Transport services.
- Added `AWSClient.createHTTPClient` which creates HTTP client based on the eventLoopGroup provided to the client. Choose `NIOTSHTTPClient` if the AWSClient has been initialised with an eventGroup of type NIOTSEventLoopGroup. Otherwise use AsyncHTTPClient.
- For signing I introduced my own [aws-signer](http://github.com/adam-fowler/aws-signer) module as it had a cleaner interface to work with than the V4 signer in aws-sdk-swift-core. 
- HTTP requests are created signed instead of creating an HTTP request and then signing it. This code has been moved to [AWSRequest](https://github.com/swift-aws/aws-sdk-swift-core/blob/6c4b21f883b4d32b8b72af6786cff0a2c35b238b/Sources/AWSSDKSwiftCore/Message/AWSRequest.swift#L91).
- Given manageCredential() was in Signers.V4 before I have moved this to [AWSClient](https://github.com/swift-aws/aws-sdk-swift-core/blob/6c4b21f883b4d32b8b72af6786cff0a2c35b238b/Sources/AWSSDKSwiftCore/AWSClient.swift#L322). The code doesn't just replace the credential it now makes a new signer. Replacing signer helped with the following.
- As the signer is the one point in the client object that can be read and written by different threads I have protected access to this with an [NSLock](https://github.com/swift-aws/aws-sdk-swift-core/blob/6c4b21f883b4d32b8b72af6786cff0a2c35b238b/Sources/AWSSDKSwiftCore/AWSClient.swift#L50). Also when a request accesses the signer it makes a copy of it and uses that copy from then on.
- Create HTTPClient at AWSClient initialisation instead of recreating for every request
- Provide the MetaDataService with the HTTPClient in AWSClient

I think that's it.

Things to do:
- At the point credentials are about to expire, multiple calls to the client could instigate multiple calls to the MetaDataService. We could probably protect against this by wrapping the signer in a `Future` and when the signer exprires the future becomes the result of the MetaDataService look up mapped to a signer. Maybe this is for another PR. 

This code has so far been tested on Mac with both HTTP clients. It has been tested with services ApiGateway, CloudFront, CognitoIdentityProvider, DynamoDB, EC2, Glacier, IAM, MediaConvert, Pinpoint, S3, SES, SNS, SQS, STS (some more comprehensively than others).


adam-fowler:
Since writing this I have done basic tests on EC2. And fixed a signing issue and the expiring credentials.

fabianfett:
I will look into this deeper. Very exciting that you have already done such a massive move into the direction of using `AsyncHTTPClient`. Thank you! 👍

I just came across:

```
            if usingNIOTransportServices {
                let request: AWSHTTPClient.Request = try createHTTPRequest(awsRequest, signer: signer)
                return invokeAWSHTTPClient(request).map { $0 }
            } else {
                let request: AsyncHTTPClient.HTTPClient.Request = try createHTTPRequest(awsRequest, signer: signer)
                return invokeAsyncHTTPClient(request).map { $0 }
            }
```

I wonder why we don't use the `AsyncHTTPClient.HTTPClient.Request` for everything?

adam-fowler:
> I wonder why we don't use the `AsyncHTTPClient.HTTPClient.Request` for everything?

It's quite an opaque type. I would have to rewrite a chunk of code in AWSHTTPClient for extracting the body. Just didn't see any worth in doing that. The request and response protocols allows for us to inject any HTTPClient into AWSClient.

fabianfett:
Had a deeper look into this and I'm really excited about your progress! 

Just as a side note I work for company that is really interested in using swift on the server and we need a great aws-sdk for this. That's why I am here and want to get involved.

> The request and response protocols allows for us to inject any HTTPClient into AWSClient.

Would you mind pointing me to where I can find these protocols?

> I would have to rewrite a chunk of code in AWSHTTPClient for extracting the body. Just didn't see any worth in doing that.

Well IMHO this has a certain code smell in the long run and is creating some code duplications where I think code duplications are not needed. Also this makes the cleanup for an `AsyncHTTPClient` that supports `NIOTS` way harder.

When I thought about creating a PR my thoughts went into this direction (see below). This is just a though map and I'm probably missing something that would prevent us from going into this direction. If you have the time I would love your feedback on this.

```swift
// create a protocol that has no type constraints. Therefore can be used as is.
protocol AWSHTTPClient {
  func execute(request: AsyncHTTPClient.HTTPClient.Request, eventLoop: AsyncHTTPClient.HTTPClient.EventLoopPreference, deadline: NIODeadline? = nil) -> EventLoopFuture<AsyncHTTPClient.HTTPClient.Response>
  func syncShutdown() throws
}

// already conforms to this
extension AsyncHTTPClient.HTTPClient: AWSHTTPClient {}

#if canImport(Networking)
class NIOTSHTTPClient: AWSHTTPClient {
  // reimplementation based on the protocol. Probably could be 
  // ton's of copy and paste from the AsyncHTTPClient just based
  // on the NIOTLSBootstrap
}
#endif

class AWSClient {
  ...
  // can be used on every request. or later once at client init.
  private func createHTTPClient(for group: EventLoopGroup) -> AWSHTTPClient {

  }
  ...
}
```

As a great side effect this should make testing of the AWSClient a breeze. Currently [the public `send` functions are untested.](https://codecov.io/gh/swift-aws/aws-sdk-swift-core/src/master/Sources/AWSSDKSwiftCore/AWSClient.swift)

fabianfett:
Also if there is something I can do to move this forward, don't hesitate to reach out!

adam-fowler:
Protocol for request creation: https://github.com/swift-aws/aws-sdk-swift-core/blob/async-http-client/Sources/AWSSDKSwiftCore/HTTP/HTTPRequest.swift
Protocol for accessing body of response: https://github.com/swift-aws/aws-sdk-swift-core/blob/6c4b21f883b4d32b8b72af6786cff0a2c35b238b/Sources/AWSSDKSwiftCore/HTTP/HTTPResponse.swift

I could add a protocol for a http client which has two associated types `request` and `response` and then we can plug in any http client.

fabianfett:
> I could add a protocol for a http client which has two associated types request and response and then we can plug in any http client.

We could do this. But in this case, if we want to keep the `HTTPClient` around for longer (as a property), we would need to have the types on the AWSClient. 

I'm still strongly in favor of keeping the httpClient around for longer (not just one request). But I don't think we want to expose the `Request` and `Response` types to the `AWSClient` consumer.

```swift
class AWSClient<Request: HTTPRequestDescription, Response: HTTPResponseDescription > {
  ...
  var httpClient: HTTPClient<Request, Response>
  ...
}
```

If you want me to I could try my best in implementing my proposed `NIOTSHTTPClient`.

adam-fowler:
Regarding keeping the client around for longer I don't think this is a huge priority. If we are supplying it with an eventLoopGroup (which we are) the initialisation/destruction is super basic and takes no time at all. Also if this is something desired I think we can cover this in a separate PR.
  
We have an HTTPClient that works with NIOTS why do we need a new one?

adam-fowler:
Ok not sure the client protocol with associated types would work. Can't pass around references to the HTTPClientProtocol (as it has associated types) and it doesn't seem possible to call functions from `some HTTPClientProtocol` with its associated types.

adam-fowler:
Regarding the NIOTSHTTPClient which takes AsyncHTTPClient Request structures and generates AsyncHTTPClient response structures. I guess you could take the current HTTPClient and fix up the `HTTPClientRequestSerializer` and `HTTPClientResponseHandler`. 

Not sure I want to tie myself to the AsyncHTTPClient structures though. Would prefer to hide them behind protocols. Whats to say the HTTPClient needs replaced again a few years down the line.

One thing for sure that could be done is reducing our current HTTPClient down to NIOTS only. It is unnecessary to support NIOSSL as we are using AsyncHTTPClient for those situations.

fabianfett:
> Not sure I want to tie myself to the AsyncHTTPClient structures though. Would prefer to hide them behind protocols. Whats to say the HTTPClient needs replaced again a few years down the line.

Highly unlikely but I think that's a fair point. 

What do you think about the following.

```swift
struct AWSRequest {
  let headers: [String: String] // incl. signed headers
  let body: ByteBuffer
}

struct AWSResponse {
  let headers: [String: String]
  let body: ByteBuffer
}

// create a protocol that has no type constraints. Therefore can be used as is.
protocol AWSHTTPClient {
  func execute(request: AWSRequest, eventLoop: EventLoopPreference, deadline: NIODeadline? = nil) -> EventLoopFuture<AWSResponse>
  func syncShutdown() throws
}

// already conforms to this
extension AsyncHTTPClient.HTTPClient: AWSHTTPClient {
  
  func execute(request: AWSRequest, eventLoop: EventLoopPreference, deadline: NIODeadline? = nil) -> EventLoopFuture<AWSResponse>
  {
    let nioRequest = request.toNIORequest()
    return self.execute(request: nioRequest, eventLoop: eventLoop, deadline: deadline)
      .map() { return $0.toAWSResponse }
  } 

}

#if canImport(Networking)
class NIOTSHTTPClient: AWSHTTPClient {
  // reimplementation based on the protocol. Probably could be 
  // ton's of copy and paste from the AsyncHTTPClient just based
  // on the NIOTLSBootstrap
}
#endif

class AWSClient {
  ...
  // can be used on every request. or later once at client init.
  private func createHTTPClient(for group: EventLoopGroup) -> AWSHTTPClient {

  }
  ...
}
```

This way we could use a protocol that doesn't need associated types and we do have clear level of separation. 

> One thing for sure that could be done is reducing our current HTTPClient down to NIOTS only. It is unnecessary to support NIOSSL as we are using AsyncHTTPClient for those situations.

I love that!

adam-fowler:
Protocol could probably work. I don't mind using NIO structures in the protocol, also request would need a method so could be
```
struct AWSRequest {
  let method: HTTPMethod
  let headers: HTTPHeaders // incl. signed headers
  let body: ByteBuffer
}
```
I'm gonna reduce AWSHTTPClient down to NIOTS first and rename it to indicate it's usage. Then we can look at this.

fabianfett:
> Regarding keeping the client around for longer I don't think this is a huge priority. If we are supplying it with an eventLoopGroup (which we are) the initialisation/destruction is super basic and takes no time at all. 

This is true for the time being. Right now the performance benefit would be just saving the malloc. But we need to consider the environment in which we want `aws-sdk-swift` to be used. If the sdk is used on a server that runs 10k req per second and each request has to make two request two DynamoDB to serve some data, this is 20k mallocs each second.

If we then consider the performance gains through connection pooling that is currently being implemented in the `AsyncHTTPClient` we compare 20k new TLS Handshakes vs. basically none since the open connections are reused over and over again. I think we look at potentially huge performance gains in this regard.

Also if this library is used on mobile the same performance gains apply, which could lead to better battery life and so on and so forth. 

> Also if this is something desired I think we can cover this in a separate PR.

Yes, I am super in favour of this. I just wanted to make sure, that we don't make future refactoring harder than it needs to be.

adam-fowler:
> Yes, I am super in favour of this. I just wanted to make sure, that we don't make future refactoring harder than it needs to be.
We can keep that in mind

fabianfett:
I really, really like where this is going!

Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift
```diff
@@ -0,0 +1,37 @@
...
+    let method: HTTPMethod
+    let headers: HTTPHeaders
+    let bodyData: Data?
+}
+
+/// HTTP Response
+protocol AWSHTTPResponse {
+    var status: HTTPResponseStatus { get }
+    var headers: HTTPHeaders { get }
+    var bodyData: Data? { get }
```
fabianfett: Just out of curiosity: Why do we use `Foundation.Data` and not `NIO.ByteBuffer`?

Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift
```diff
@@ -0,0 +1,37 @@
+//
```
fabianfett: Does it make sense to rename this file to `AWSHTTPClient`?

Sources/AWSSDKSwiftCore/HTTP/NIOTSHTTPClient.swift
```diff
@@ -9,6 +9,7 @@
 // and the swift-server's swift-nio-http-client
 // https://github.com/swift-server/swift-nio-http-client
 //
+#if canImport(Network)
```
fabianfett: Sadly I can't comment on line 20. The `#if canImport(Network)` can be removed.

Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift
```diff
@@ -0,0 +1,37 @@
...
+/// HTTP Request
+struct AWSHTTPRequest {
+    let url: URL
+    let method: HTTPMethod
+    let headers: HTTPHeaders
+    let bodyData: Data?
+}
+
+/// HTTP Response
+protocol AWSHTTPResponse {
```
fabianfett: I really like that the response is a protocol! 🚀

Package.swift
```diff
@@ -11,13 +11,17 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio.git", .upToNextMajor(from:"2.8.0")),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", .upToNextMajor(from:"2.4.0")),
         .package(url: "https://github.com/apple/swift-nio-transport-services.git", .upToNextMajor(from:"1.0.0")),
+        .package(url: "https://github.com/swift-server/async-http-client.git", .upToNextMajor(from:"1.0.0")),
+        .package(url: "https://github.com/adam-fowler/aws-signer.git", .upToNextMajor(from:"1.0.0")),
```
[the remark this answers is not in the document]
adam-fowler: It's a lib I had already written. If we stick with it can merge into aws-sdk-swift-core.

Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift
adam-fowler: We were already generating `Data` objects for requests. If we were to move to`NIO.ByteBuffer` I would want to make sure we aren't making additional copies of the body data when moving between data types. I just haven't had time to look into that.

Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift
adam-fowler: yep

Sources/AWSSDKSwiftCore/HTTP/NIOTSHTTPClient.swift
adam-fowler: Yep should be removed, thought I had cleaned up those imports already. Oh well will do it again

Package.swift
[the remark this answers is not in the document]
fabianfett: Sound awesome!

Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift
fabianfett: Okay. No problem at all!

Sources/AWSSDKSwiftCore/HTTP/NIOTSHTTPClient.swift
adam-fowler: sorted

Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift
adam-fowler: sorted

Sources/AWSSDKSwiftCore/HTTP/HTTPClient.swift
adam-fowler: done now

adam-fowler:
I think this PR is ready to merge. Guys can I get a review.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -15,8 +17,32 @@ import NIOTransportServices
 /// Convenience shorthand for `EventLoopFuture`.
 public typealias Future = EventLoopFuture
 
+/// Helper struct for ensuring atomic access to a structure
+struct AtomicProperty<T> {
```
fabianfett: I don't like this here. Maybe we can work around this with the approach in: https://github.com/adam-fowler/aws-signer/pull/2

Sources/AWSSDKSwiftCore/HTTP/AWSHTTPClient.swift
```diff
@@ -0,0 +1,37 @@
...
+/// Protocol defining requirements for a HTTPClient
+protocol AWSHTTPClient {
+    /// Execute HTTP request and return a future holding a HTTP Response
+    func execute(request: AWSHTTPRequest, timeout: TimeAmount) -> Future<AWSHTTPResponse>
+    
+    /// This should be called before an HTTP Client can be de-initialised
+    func syncShutdown() throws
+    
+    /// Event loop group used by client
+    var eventLoopGroup: EventLoopGroup {get}
```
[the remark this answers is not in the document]
adam-fowler: sorted

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
```diff
@@ -0,0 +1,23 @@
+import AsyncHTTPClient
+import Foundation
+import NIO
+
+/// comply with AWSHTTPClient protocol
+extension AsyncHTTPClient.HTTPClient: AWSHTTPClient {
+    func execute(request: AWSHTTPRequest, timeout: TimeAmount) -> EventLoopFuture<AWSHTTPResponse> {
+        let requestBody: AsyncHTTPClient.HTTPClient.Body?
+        if let body = request.body {
+            requestBody = AsyncHTTPClient.HTTPClient.Body.byteBuffer(body)
```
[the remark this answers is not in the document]
adam-fowler: sorted

Sources/AWSSDKSwiftCore/HTTP/NIOTSHTTPClient.swift
```diff
@@ -218,16 +151,16 @@ public final class HTTPClient {
             head.headers.replaceOrAdd(name: "Host", value: hostname)
             head.headers.replaceOrAdd(name: "User-Agent", value: "AWS SDK Swift Core")
             head.headers.replaceOrAdd(name: "Accept", value: "*/*")
-            head.headers.replaceOrAdd(name: "Content-Length", value: request.body.count.description)
+            if let body = request.body {
+                head.headers.replaceOrAdd(name: "Content-Length", value: body.readableBytes.description)
+            }
             // TODO implement keep-alive
```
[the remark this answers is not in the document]
adam-fowler: removed

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
```diff
@@ -26,26 +26,27 @@ public struct AWSResponse {
...
-        for (key, value) in response.head.headers {
+        for (key, value) in response.headers {
             responseHeaders[key.description] = value
         }
 
         return responseHeaders
     }
     
-    private static func createBody(from response: HTTPClient.Response, serviceProtocolType: ServiceProtocolType, raw: Bool) throws -> Body {
+    private static func createBody(from response: AWSHTTPResponse, serviceProtocolType: ServiceProtocolType, raw: Bool) throws -> Body {
```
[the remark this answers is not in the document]
adam-fowler: Lets look at this for another PR

Sources/AWSSDKSwiftCore/Credential/Credential.swift
```diff
@@ -45,6 +29,15 @@ public struct Credential: CredentialProvider {
         self.sessionToken = sessionToken ?? ProcessInfo.processInfo.environment["AWS_SESSION_TOKEN"]
         self.expiration = expiration
     }
+    
+    func nearExpiration() -> Bool {
+        if let expiration = self.expiration {
+            // are we within 5 minutes of expiration?
+            return Date().addingTimeInterval(5.0 * 60.0) > expiration
+        } else {
```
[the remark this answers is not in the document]
adam-fowler: gonna leave it

Sources/AWSSDKSwiftCore/AWSClient.swift
adam-fowler: I understand your concerns about this, but I think the changes you have suggested are too big. See my version in the comments of your PR.

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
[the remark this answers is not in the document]
fabianfett: Track issue here: https://github.com/swift-aws/aws-sdk-swift-core/issues/161

fabianfett:
@adam-fowler would you mind committing your AWSSigner library into this pr?

A couple of changes I would propose to the `AWSSigner` lib.

1. Naming. I think a `Credential` should be a protocol that have the properties `accessKeyId`, `accessKeyId`, `secretAccessKey`. An implementation of this could be a `StaticCredential`. 

2. Credentials should be injected into the signer at every request. Therefore the signer becomes stateless and can stay as a property on the `AWSClient` for the complete lifecycle.

The changes are already implemented here with the tests also running:

https://github.com/adam-fowler/aws-signer/pull/2/files

fabianfett:
I like the shortcut you propose at https://github.com/swift-aws/aws-sdk-swift-core/pull/160#issuecomment-562098040.

Though I would like to underline a couple of naming conventions, that I really would like to see:

1. Something that serves/gives a `Credential` is a `CredentialProvider`. I stay away as far as possible from names with `Future` in it. For me `Future`s are always typealiases to some `EventLoopFuture`.

```
/// Protocol for providing credential details for accessing AWS services
public protocol CredentialProvider {
    func getCredential() -> EventLoopFuture<Credential>
}
```

2. Separating `CredentialProvider` and `Signing`

> Once you have your credential you can then create an AWSSigner with that and sign your request. I'm not a bug fan of the splitting of AWSSigner into credential data and the code that uses that credential data.

Why are you not a big fan of this? I'm pretty sure the `CredentialProvider` part of this is nowhere reusable. The `AWSSigner` part is (that's why I would try to keep it as NIO free as possible). I basically have the same implementation in a couple of iOS projects that need to sign requests against APIGateway. Using a protocol to inject the `Credential` makes it super easy for a consumer to use it. The `AWSClient`s task is then to put the pieces together. IMHO this a super clear separation of concerns. 

3. Initializing the `CredentialProvider` in the `init` and keeping it as a property for the complete lifetime of the `AWSClient`.

adam-fowler:
> @adam-fowler would you mind committing your AWSSigner library into this pr?
> 
> A couple of changes I would propose to the `AWSSigner` lib.
> 
>     1. Naming. I think a `Credential` should be a protocol that have the properties `accessKeyId`, `accessKeyId`, `secretAccessKey`. An implementation of this could be a `StaticCredential`.
> 
>     2. Credentials should be injected into the signer at every request. Therefore the signer becomes stateless and can stay as a property on the `AWSClient` for the complete lifecycle.
> 
> 
> The changes are already implemented here with the tests also running:
> 
> https://github.com/adam-fowler/aws-signer/pull/2/files

I'm just in the process of doing this. While I'm not that bothered about the naming. Not so sure about the second point. The signer doesn't become stateless as it still has the signing name and region. I'd prefer we just create a new signer, with each request. It seems a little cleaner.

fabianfett:
> I'd prefer we just create a new signer, with each request. It seems a little cleaner.

If we make it a struct I'm all for it! There is no need for the signer to be a class.

adam-fowler:
> If we make it a struct I'm all for it! There is no need for the signer to be a class.
It was a class because of previous requirements which are not valid anymore

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -131,36 +141,67 @@ public class AWSClient {
         case .useAWSClientShared:
             self.eventLoopGroup = AWSClient.sharedEventLoopGroup
         }
+        self.httpClient = AWSClient.createHTTPClient(eventLoopGroup: self.eventLoopGroup)
```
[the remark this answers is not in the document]
adam-fowler: Given we prefix most of the values below with `self.` I would say that is consistent with the rest of the function. I think this is fine within an init() function.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -131,36 +141,67 @@ public class AWSClient {
...
-        let futureResponse = client.connect(nioRequest)
 
-        futureResponse.whenComplete { _ in
-            do {
-                try client.syncShutdown()
-            } catch {
-                print("Error closing connection: \(error)")
-            }
-        }
+    /// invoke AWS request, create HTTP request from AWS request and then make request. Return response. Function chooses which HTTP client to use based
```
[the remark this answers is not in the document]
adam-fowler: removed

Sources/AWSSigner/hash.swift
```diff
@@ -0,0 +1,78 @@
+//
```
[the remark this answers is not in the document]
adam-fowler: We can remove most of these from AWSSDKSwiftCore. MD5 is used in S3 middleware. Also one of the SHA256 functions is used in Glacier middleware. I've removed HMAC.swift as it isn't needed anymore.

Sources/AWSSDKSwiftCore/AWSClient.swift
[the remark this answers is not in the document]
jonnymacs: hmm, not exactly. The `self.` prefix is required on the assignment (left) side and we use that appropriately to assign values below. But it's not required when reading the value of the `eventLoopGroup` instance constant. I'm suggesting to remove `self.` from where we read that value.

So it would be

`self.httpClient = AWSClient.createHTTPClient(eventLoopGroup: eventLoopGroup)`

If that makes sense.

Then to disambiguate above when we are first assigning that constant, it would be better to name the unwrapped EventLoopGroup something that indicates it's not the local instance var. So for that I'm recommending `sharedEventLoopGroup`

Let me know if this still doesn't make sense. I can explain over discord.

And goes without saying - these are nit picks and not a big deal. Just things that I notice when I'm reading the code that imo would make it more clear what's happening.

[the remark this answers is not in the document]
adam-fowler: Sorry missed which bit you were talking about. Removed `self` and renamed `eventLoopGroup` in the switch statement

Sources/AWSSigner/hash.swift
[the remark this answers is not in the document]
adam-fowler: Removed all hashing functions from AWSSDKSwiftCore as they aren't needed. Added a couple that AWSSDKSwift services needed (eg md5). Also added `@_exported` to AWSSigner and NIO imports so their symbols are available if AWSSDKSwiftCore is imported.

Sources/AWSSDKSwiftCore/Exports.swift
```diff
@@ -0,0 +1,2 @@
+@_exported import AWSSignerV4
+@_exported import NIO
```
fabianfett: I'm not sure if we really want to expose NIO?

Sources/AWSSDKSwiftCore/HTTP/NIOTSHTTPClient.swift
```diff
@@ -0,0 +1,225 @@
...
+                        return channel.pipeline.addHandlers(handlers)
+                }
+            }
+            .connect(host: hostname, port: port)
+            .flatMap { channel -> EventLoopFuture<Void> in
+                return channel.writeAndFlush(request)
+            }
+            .whenFailure { error in
+                response.fail(error)
+        }
```
fabianfett: `}` should be indented.

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
```diff
@@ -103,24 +115,26 @@ public struct AWSRequest {
         }
         return awsRequest
     }
+}
+
 
-    // Convert string to NIO HTTP Method
-    fileprivate func nioHTTPMethod(from: String) -> HTTPMethod {
+extension HTTPMethod {
```
fabianfett: AFAIK, we can use the rawValue directly from SwiftNIO. Was released with `2.13.0`: https://github.com/apple/swift-nio/pull/1329

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
```diff
@@ -25,26 +25,27 @@ public struct AWSResponse {
...
-    init(from response: HTTPClient.Response, serviceProtocolType: ServiceProtocolType, raw: Bool = false) throws {
-        self.status = response.head.status
+    init(from response: AWSHTTPResponse, serviceProtocolType: ServiceProtocolType, raw: Bool = false) throws {
+        self.status = response.status
         self.headers = AWSResponse.createHeaders(from: response)
         self.body = try AWSResponse.createBody(from: response, serviceProtocolType: serviceProtocolType, raw: raw)
     }
 
-    private static func createHeaders(from response: HTTPClient.Response) -> [String: String] {
+    private static func createHeaders(from response: AWSHTTPResponse) -> [String: String] {
```
fabianfett: this method looks unswifty:
```
  private func getHeaders() -> [String: String]
```

would probably address this way better.

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
```diff
@@ -25,26 +25,27 @@ public struct AWSResponse {
...
-        for (key, value) in response.head.headers {
+        for (key, value) in response.headers {
             responseHeaders[key.description] = value
         }
 
         return responseHeaders
     }
     
-    private static func createBody(from response: HTTPClient.Response, serviceProtocolType: ServiceProtocolType, raw: Bool) throws -> Body {
+    private static func createBody(from response: AWSHTTPResponse, serviceProtocolType: ServiceProtocolType, raw: Bool) throws -> Body {
```
fabianfett: Same as above applies.

Sources/AWSSDKSwiftCore/Message/Body.swift
```diff
@@ -77,4 +77,41 @@ extension Body {
...
+            buffer.writeString(text)
+            return buffer
+
+        case .buffer(let data):
+            var buffer = ByteBufferAllocator().buffer(capacity: data.count)
+            buffer.writeBytes(data)
+            return buffer
+
+        case .json(let data):
+            if data.isEmpty {
```
fabianfett:
```
guard !data.isEmpty else {
  return nil
}

// continue with no indent.
```

Sources/AWSSignerV4/credentials.swift
```diff
@@ -0,0 +1,47 @@
+//
+//  credentials.swift
+//  aws-sign
+//
+//  Created by Adam Fowler on 29/08/2019.
+//
+import Foundation
```
fabianfett: Specialise on Foundation.ProcessInfo

Sources/AWSSignerV4/hash.swift
```diff
@@ -0,0 +1,167 @@
+//
+//  hash.swift
+//  AsyncHTTPClient
+//
+//  Created by Adam Fowler on 2019/08/29.
+//
+
+import Foundation
```
fabianfett: import detail...

Sources/AWSSignerV4/credentials.swift
```diff
@@ -0,0 +1,47 @@
...
+
+/// Protocol for providing credential details for accessing AWS services
+public protocol Credential {
+    var accessKeyId: String {get}
+    var secretAccessKey: String {get}
+    var sessionToken: String? {get}
+}
+
+/// basic version of Credential where you supply the credentials
+public struct StaticCredential: Credential {
```
fabianfett: We have `StaticCredential` as well as `EnvironmentCredential` in AWSSDKCore as well. I'd remove it from here.

Sources/AWSSignerV4/signer.swift
```diff
@@ -0,0 +1,233 @@
+//
+//  signer.swift
+//  AWSSigner
+//
+//  Created by Adam Fowler on 2019/08/29.
+//  Amazon Web Services V4 Signer
+//  AWS documentation about signing requests is here https://docs.aws.amazon.com/general/latest/gr/signing_aws_api_requests.html
+//
+
+import Foundation
```
fabianfett: import specific Foundation structs/classes

Tests/AWSSDKSwiftCoreTests/SignersV4Tests.swift
```diff
@@ -8,7 +8,7 @@
 
 import XCTest
 @testable import AWSSDKSwiftCore
-
+/*
```
fabianfett: remove completely.

Sources/AWSSignerV4/signer.swift
```diff
@@ -0,0 +1,233 @@
...
+        query += "&X-Amz-Signature=\(signature(signingData: signingData))"
+        
+        // Add signature to query items and build a new Request
+        let signedURL = URL(string: url.absoluteString.split(separator: "?")[0]+"?"+query)!
+
+        return signedURL
+    }
+    
+    /// structure used to store data used throughout the signing process
+    class SigningData {
```
fabianfett: having this as a `class` looks wrong.

Sources/AWSSDKSwiftCore/Exports.swift
adam-fowler: Pretty anytime someone is interacting with AWSSDKSwift they are required to deal with EventLoopFutures which requires them to `import NIO`. We could expose selected objects like EventLoopFuture, EventLoopGroup, ByteBuffer maybe

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
adam-fowler: Can't call a method on a struct that isn't fully initialised, that is why it is a static method, happy to rename though

Sources/AWSSDKSwiftCore/Message/Body.swift
adam-fowler: Not very sure why this code is even in there. I can't see a reason how someone could initialize json with an empty body. Also the test is only there for json. I have another branch where I am working on Body can we look into this then when I put it up (next few days)

Sources/AWSSignerV4/hash.swift
adam-fowler: This is all going in the bin in the near future, gonna leave it as is just now

Sources/AWSSignerV4/credentials.swift
adam-fowler: No we don't. AWSSDKSwiftCore gets those from AWSSignerV4

Sources/AWSSignerV4/signer.swift
adam-fowler: done

Tests/AWSSDKSwiftCoreTests/SignersV4Tests.swift
adam-fowler: done

Sources/AWSSignerV4/signer.swift
adam-fowler: ¯\_(ツ)_/¯
done

Sources/AWSSignerV4/credentials.swift
adam-fowler: done

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
adam-fowler: done, plus upped the swift-nio version requirement in Package.swift

Sources/AWSSDKSwiftCore/HTTP/NIOTSHTTPClient.swift
adam-fowler: whole function was indented weird, fixed

Sources/AWSSignerV4/credentials.swift
fabianfett: right you are. sorry.

Sources/AWSSDKSwiftCore/Exports.swift
adam-fowler: See Changes, just exporting the objects we expose in our public APIs

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
fabianfett: Why do we still have this method here? Don't we want to use `HTTPMethod(rawValue: String)` directly?

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
fabianfett: both methods are not really large, what do you think about moving the logic directly into the initialiser?

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
fabianfett: both methods are not really large, what do you think about moving the logic directly into the initialiser?

Sources/AWSSDKSwiftCore/Message/Body.swift
fabianfett: no fix for now.

Sources/AWSSignerV4/hash.swift
fabianfett: What do you mean with "in the bin"? Would love to keep this tidy... Just `import Foundation` is a regression imho.

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
adam-fowler: > Why do we still have this method here? Don't we want to use `HTTPMethod(rawValue: String)` directly?

We are using it. I ditched the old conversion method so we are forced to. Maybe github has got a little confused. Checkout the actual file

Sources/AWSSignerV4/hash.swift
adam-fowler: All of hash.swift is about to replaced by swift-crypto

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
adam-fowler: ok

adam-fowler: done

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
adam-fowler: see above

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
```diff
@@ -24,33 +24,31 @@ public struct AWSResponse {
...
-
-    private static func createHeaders(from response: HTTPClient.Response) -> [String: String] {
+    init(from response: AWSHTTPResponse, serviceProtocolType: ServiceProtocolType, raw: Bool = false) throws {
+        self.status = response.status
+        
+        // headers
         var responseHeaders: [String: String] = [:]
-        for (key, value) in response.head.headers {
+        for (key, value) in response.headers {
             responseHeaders[key.description] = value
```
fabianfett: `key` is a string already.

Sources/AWSSignerV4/signer.swift
fabianfett: IMHO this is a value type that wraps a number of properties.
Considering performance: This is a type that is currently created on every request, this malloc could easily be removed.

adam-fowler: The allocation would be a reason to change it to a struct but unfortunately the object contains a dictionary which causes an allocation everytime we copy it (or pass it down through a function).

adam-fowler: The additional allocation would be an argument to make it a struct. I'm not certain the dictionary in SigningData doesn't cause additional allocations on each copy though. I'll make it a struct in this case, but we need good reasons for choosing either a struct/class. In the case of read-only objects neither struct or class are necessarily the right choice. If you look at the guidelines [here](https://developer.apple.com/swift/blog/?id=10) a read-only object is not really covered by any of the points on which to choose. We can carry this conversation elsewhere though.

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
adam-fowler: done

PR #144, opened 2019-11-11
Adjustments as discussed in PR #139

fabianfett:
- `AWSClient`’s static shared instance of `eventLoopGroup` is now private.
- `AWSClient`’s initialiser does not have a default property for `EventLoopGroupProvider` anymore.


adam-fowler:
Did you mean to set eventLoopGroup private? Thought we were setting sharedEventLoopGroup to private.

adam-fowler:
All good

PR #145, opened 2019-11-14
GitHub actions

adam-fowler:
Integration with github actions.
Compared to Travis it is faster, but the UI seems a little clunky at the moment. 
Thought we could have both running for a while to see which works better.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -1,0 +1,38 @@
+name: Swift
+
+on: [push]
+
+jobs:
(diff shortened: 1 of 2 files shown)


adam-fowler:
Just noticed on pull requests it doesn't do a rebuild.

fabianfett:
Hi, we don't even need to pull the image ourselfs.

fabianfett:
I just created a PR with my comments. #146 Maybe you like them. ;)

PR #147, opened 2019-11-19
Improvements to GitHub Actions

fabianfett:
- [x] builds on pr
- [x] tests swift on linux versions with matrix builds
- [x] uploads codecov from linux as well (>= 5.1)
- [x] uses linux auto test discovery (>= 5.1)

What else do we want to achieve with this pr?

- [ ] Remove travis?

Looking forward to your feedback: @adam-fowler @Yasumoto @jonnymacs

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -1,40 +1,62 @@
 name: Swift
 
-on: [push]
-
-env:
-   CODECOV_TOKEN: ${{ secrets. CODECOV_TOKEN }}
+on: 
+  push:
+    branches: 
+    - master
+  pull_request:
+    branches: 
+    - master
 
 jobs:
   macos:
     runs-on: macOS-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: Build
+    - name: Checkout
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: SPM build
       run: swift build
-    - name: Run tests
+    - name: SPM tests
       run: swift test
-    - name: Code Coverage
-      run:
-        swift package generate-xcodeproj &&
-        xcodebuild -quiet -parallel-testing-enabled YES -scheme AWSSDKSwiftCore-Package -enableCodeCoverage YES build test &&
-        bash <(curl -s https://codecov.io/bash) -J 'AWSSDKSwiftCore' 
-            
-  linux-swift-5-0:
-    runs-on: ubuntu-16.04
-    steps:
-    - uses: actions/checkout@v1
-    - name: Install
-      run: docker pull swift:5.0
-    - name: Build & Test
-      run: docker run -v "$(pwd):/root/" swift:5.0 /bin/sh -c "apt-get -qq update && apt-get -q -y install tzdata libssl-dev zlib1g-dev && cd /root/ && swift build && swift test --parallel"
-      
-  linux-swift-5-1-1:
-    runs-on: ubuntu-16.04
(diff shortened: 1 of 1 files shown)


.github/workflows/swift.yml
```diff
@@ -1,40 +1,62 @@
 name: Swift
 
-on: [push]
-
-env:
-   CODECOV_TOKEN: ${{ secrets. CODECOV_TOKEN }}
+on: 
+  push:
+    branches: 
+    - master
```
adam-fowler: why just master?

.github/workflows/swift.yml
```diff
@@ -1,40 +1,62 @@
...
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: Install dependencies
+      run: |
+        apt-get update -qq
+        apt-get install -q -y tzdata libssl-dev zlib1g-dev
+    - name: Test
+      run: swift test --parallel --enable-code-coverage
+      if: matrix.tag < 5.1
```
adam-fowler: Can we swap the if: and run: elements around. It just reads odd with the if after the run

.github/workflows/swift.yml
fabianfett: Because otherwise ci is automatically runs twice on each branch, that has a pull request, which IMHO is a waist of resources – even if it's free. 

I guess the normal way is as follows: one branch where the magic happens (master in this case). All other branches want to merge into this branch sooner or later. If one needs ci, one can always create a "Draft Request" for a branch that isn't fully baked and ready for a real PR.

.github/workflows/swift.yml
fabianfett: fixed.

.github/workflows/swift.yml
adam-fowler: I'd much prefer we had ci on all branches though. It feels like I'm losing something here. Creating a PR just so you can get a ci build done seems to send the wrong messages about what a PR is.

fabianfett: My idea was to minimize the number of runs the ci does, that are totally wasted resources. I have seen in the previous prs, travis has always run the exact test twice. If there is a branch that needs CI on every push, one can just add the branch to the ci file on that branch.

With that we are resourcefulness in the normal case but also have the freedom for more ci on push if needed. But in this case it's an opt in, whereas a wildcard branch doesn't even has the chance to opt out.

jonnymacs: I don't have a strong opinion here, so I think a tie should go to the incumbent :-)

I see the logic and appreciate the noble pursuit to be resourceful. I really wouldn't be too bothered if we went ahead with ci on PRs only. But I do think it's a nice convenience to have ci on each branch as a default. I find we most often test locally and push as infrequently as possible to branches on the remote repo for ci. So it's really not that much waste imo.

fabianfett: fixed.

adam-fowler:
Now we need to do this for aws-sdk-swift :-)

fabianfett:
@adam-fowler If that's okay for you I would provide a PR for `aws-sdk-swift` next week.

adam-fowler:
@fabianfett sounds good. You gonna merge this one in?

fabianfett:
@adam-fowler May I?

adam-fowler:
@fabianfett go for it

adam-fowler:
Err ok too late. Maybe for the next one.

PR #150, opened 2019-11-24
If GET request has a header then sign request using headers

adam-fowler:
If a GET request has headers sign the request with an authorization header. If we don't do this AWS complains that the headers are not signed.

Fixes issue sending s3.getObjectAcl() request.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -334,6 +334,9 @@ extension AWSClient {
             case .restjson:
                 return try createNIORequestWithSignedHeader(awsRequest)
             default:
+                if awsRequest.httpHeaders.count > 0 {
+                    return try createNIORequestWithSignedHeader(awsRequest)
+                }
                 return try createNIORequestWithSignedURL(awsRequest)
             }
         default:


Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -334,6 +334,9 @@ extension AWSClient {
             case .restjson:
```
fabianfett: I, personally, don't like switches within switches? Maybe this is a chance for:

```swift
switch (awsRequest.httpMethod, self.serviceProtocol.type, awsRequest.httpHeaders.count) {
case ("GET",  .restjson, _), 
     ("HEAD", .restjson, _):
    return try createNIORequestWithSignedHeader(awsRequest)
case ("GET",  _, 0),
     ("HEAD", _, 0):
    return try createNIORequestWithSignedURL(awsRequest)
case ("GET",  _, 1...),
     ("HEAD", _, 1...):
    return try createNIORequestWithSignedHeader(awsRequest)
}
```

adam-fowler: Not totally keen on that. Combining the httpMethod and type I'm ok with but including the header count just seems wrong. 
```swift
switch (awsRequest.httpMethod, self.serviceProtocol.type) {
case ("GET",  .restjson), ("HEAD", .restjson):
    return try createNIORequestWithSignedHeader(awsRequest)
case ("GET",  _), ("HEAD", _):
    if awsRequest.httpHeaders.count > 0 {
        return try createNIORequestWithSignedHeader(awsRequest)
    }
    return try createNIORequestWithSignedURL(awsRequest)
default:
    return try createNIORequestWithSignedHeader(awsRequest)
}
```
Also will that mean the switch tests for the header count range twice?

fabianfett: I can live with that totally fine!

jonnymacs:
@adam-fowler This change is throwing an alarm for me.

Earlier in the year I battled S3 running in a container with IAM ecs credentials. HEAD requests were being denied as unauthorized when sent with signed headers. Locally everything worked, but when the token was introduced in the IAM role credential env the calls failed. Only by [switching all HEAD and requests to the signedURL](https://github.com/swift-aws/aws-sdk-swift-core/commit/ebbf137f5f1c04b4d9a44f46604a961d619245bf) method did it work.

I assume GETs would suffer the same fate as even on a standard getObject / headObject request there are headers set

```
        // set x-amz-target header
        if let target = amzTarget {
            headers["x-amz-target"] = "\(target).\(operationName)"
        }
```

So I *think* with this code change, we will trigger use of signedHeaders again for these calls and I these calls will begin to fail again in my ecs env.

I haven't tested with this PR but that's my prediction. If you want to try setting up an env with IAM credentials to test that would be great. I don't think I will have time this week (day job / thanksgiving stuff) but can test it next week.

adam-fowler:
> @adam-fowler This change is throwing an alarm for me.
> 
> Earlier in the year I battled S3 running in a container with IAM ecs credentials. HEAD requests were being denied as unauthorized when sent with signed headers. Locally everything worked, but when the token was introduced in the IAM role credential env the calls failed. Only by [switching all HEAD and requests to the signedURL](https://github.com/swift-aws/aws-sdk-swift-core/commit/ebbf137f5f1c04b4d9a44f46604a961d619245bf) method did it work.
> 
> I assume GETs would suffer the same fate as even on a standard getObject / headObject request there are headers set
> 
> ```
>         // set x-amz-target header
>         if let target = amzTarget {
>             headers["x-amz-target"] = "\(target).\(operationName)"
>         }
> ```
> 
> So I _think_ with this code change, we will trigger use of signedHeaders again for these calls and these calls will begin to fail again in my ecs env.
> 
> I haven't tested with this PR but that's my prediction. If you want to try setting up an env with IAM credentials to test that would be great. I don't think I will have time this week (day job / thanksgiving stuff) but can test it next week.

Did a quick check. There isn't a single service that sets "amzTarget" and also uses GET or HEAD. I will do a quick EC2 test as well. Interested as to why it works locally but not on an EC2 instance.

adam-fowler:
@jonnymacs verified on EC2 instance. Tested headObject, getObject, listBuckets, listObjects

adam-fowler:
whoops sorry @fabianfett forgot to stick the switch changes in. Will do that in separate PR

PR #151, opened 2019-11-24
Allow xml decoder to decode attributes as members

### LINKED ISSUE #149: S3.GetObjectAcl throws decoding error
**Describe the bug**
S3.GetObjectAcl throws decoding error. It occurs because one of the required parameters for the output shape is stored as an xsi xml attribute which we don't parse.

**To Reproduce**
```
let request = S3.GetObjectAclRequest(bucket: testData.bucket, key: testData.key, requestPayer: .requester)
let response = try client.getObjectAcl(request).wait()
```

adam-fowler:
Needed to change XMLDecoder() to work with XML.Node instead of XML.Element internally.

This fixed #149

### DIFF
--- a/Sources/AWSSDKSwiftCore/Encoder/XMLDecoder.swift
+++ b/Sources/AWSSDKSwiftCore/Encoder/XMLDecoder.swift
@@ -236,18 +236,17 @@ fileprivate class _XMLDecoder : Decoder {
                         element = parent
                     }
                 }
-                if let entryName = entryName {
-                    if let entries = (element as? XML.Element)?.elements(forName: entryName) {
-                        for entry in entries {
-                            if let keyNode = entry.child(for: keyName), let valueNode = entry.child(for: valueName) {
-                                guard let keyString = keyNode.stringValue else { continue }
-                                if let key = Key(stringValue: keyString) {
-                                    allKeys.append(key)
-                                    // store value elements for later
-                                    allValueNodes[keyString] = valueNode
-                                }
-                            }
+                if let entryName = entryName, let entries = (element as? XML.Element)?.elements(forName: entryName) {
+                    for entry in entries {
+                        guard let keyNode = entry.child(for: keyName),
+                            let valueNode = entry.child(for: valueName),
+                            let keyString = keyNode.stringValue,
+                            let key = Key(stringValue: keyString) else {
+                            continue
                         }
+                        allKeys.append(key)
+                        // store value elements for later
+                        allValueNodes[keyString] = valueNode
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


Sources/AWSSDKSwiftCore/Encoder/XMLDecoder.swift
```diff
@@ -261,30 +268,32 @@ fileprivate class _XMLDecoder : Decoder {
...
             if expandedDictionary {
-                guard let child = allValueElements[key.stringValue] else {
+                guard let child = allValueNodes[key.stringValue] else {
                     throw DecodingError.keyNotFound(key, DecodingError.Context(codingPath: codingPath, debugDescription:"Failed to find key value in expanded dictionary. Should not get here"))
                 }
                 return child
             } else {
-                guard let child = element.child(for: key) else {
-                    throw DecodingError.keyNotFound(key, DecodingError.Context(codingPath: codingPath, debugDescription: "Key not found"))
+                if let child = element.child(for: key) {
```
fabianfett: If I see this correctly we could put this one level up:

```swift
func child(for key: CodingKey) throws -> XML.Node {
  if expandedDictionary {
    guard let child = allValueElements[key.stringValue] else {
      throw DecodingError.keyNotFound(key, DecodingError.Context(codingPath: codingPath, debugDescription:"Failed to find key value in expanded dictionary. Should not get here"))
    }
    return child
  } 
  else if let child = element.child(for: key) {
    return child
  }
  else if let attribute = (element as? XML.Element)?.attribute(for: key) {
    return attribute
  }
  throw DecodingError.keyNotFound(key, DecodingError.Context(codingPath: codingPath, debugDescription: "Key not found"))
}
```

```

File: Sources/AWSSDKSwiftCore/Encoder/XMLDecoder.swift [hunk 2]
```diff
@@ -463,10 +470,10 @@ fileprivate class _XMLDecoder : Decoder {
             if let elements = elements {
                 self.elements = elements
             } else {
-                if let parent = element.parent as? XML.Element {
+                if let parent = element.parent {
```
fabianfett: same as above would be in favor off moving this code one level to the left ;)

Sources/AWSSDKSwiftCore/Encoder/XMLDecoder.swift
```diff
@@ -224,21 +230,22 @@ fileprivate class _XMLDecoder : Decoder {
...
                 if let entryName = entryName {
-                    let entries = element.elements(forName: entryName)
-                    for entry in entries {
-                        if let keyElement = entry.child(for: keyName), let valueElement = entry.child(for: valueName) {
-                            guard let keyString = keyElement.stringValue else { continue }
-                            if let key = Key(stringValue: keyString) {
-                                allKeys.append(key)
-                                // store value elements for later
-                                allValueElements[keyString] = valueElement
+                    if let entries = (element as? XML.Element)?.elements(forName: entryName) {
```
fabianfett: Maybe combine if?

```swift
if let entryName = entryName, let entries = (element as? XML.Element)?.elements(forName: entryName) {
```

Sources/AWSSDKSwiftCore/Encoder/XMLDecoder.swift
```diff
@@ -224,21 +230,22 @@ fileprivate class _XMLDecoder : Decoder {
...
+                        for entry in entries {
+                            if let keyNode = entry.child(for: keyName), let valueNode = entry.child(for: valueName) {
+                                guard let keyString = keyNode.stringValue else { continue }
+                                if let key = Key(stringValue: keyString) {
```
fabianfett: Maybe combine `if` and `guard` and `if`?
```swift
guard let keyNode = entry.child(for: keyName), 
      let valueNode = entry.child(for: valueName), 
      let keyString = keyNode.stringValue, 
      let key = Key(stringValue: keyString) else 
{
  continue
}
allKeys.append(key)
// store value elements for later
allValueNodes[keyString] = valueNode
```

IMHO this would make reading way easier!

adam-fowler: These all make sense

PR #154, opened 2019-11-26
Use AWSClient.eventLoopGroup in HTTPClient (merge into v3.x)

adam-fowler:
- HTTPClient has to be provided with an eventLoopGroup. It cannot create its own. Therefore does not need a close() function
- Use the static AWSClient.eventGroup in HTTPClient

This fixes the metadataservice hangs being experienced just now

We could extend this to be similar to the eventLoopGroupProvider changes in v4.x, but that would need considerable more changes on the aws-sdk-swift side as well


adam-fowler:
Tested on mac and ec2

PR #156, opened 2019-11-27
Use github actions to generate documentation on release

adam-fowler:


.github/workflows/gen-docs.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Install Jazzy
+      run: gem install jazzy
+    - name: Build
+      run: |
+        swift package generate-xcodeproj
+        jazzy --clean
+    - name: Commit
+      run: |
+        remote_repo="https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+        remote_branch="gh-pages"
```
fabianfett: Maybe we do env variables in all CAPS?
Further we could define env variables in the env section.

```yaml
- name: Commit
  env:
    REMOTE_REPO: "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
    REMOTE_BRANCH: "gh-pages"
  run: |
    ...
```

.github/workflows/gen-docs.yml
```diff
@@ -0,0 +1,36 @@
+name: Generate Documentation
+
+on: [release]
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
fabianfett: Just making sure, since I can't look into the repos settings. We use the default `GITHUB_TOKEN`? As described [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token#about-the-github_token-secret). No extra GITHUB_TOKEN has been set in the settings?

.github/workflows/gen-docs.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  build:
+
+    runs-on: macOS-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
fabianfett: Maybe we want to name this one as well? "Checkout"?

.github/workflows/gen-docs.yml
adam-fowler: this the default GITHUB_TOKEN

.github/workflows/gen-docs.yml
adam-fowler: ok

.github/workflows/gen-docs.yml
adam-fowler: done, plus moved GITHUB_TOKEN here

adam-fowler: Tried moving the environment vars to env: section. It doesn't like expanding environment vars in that section so $REMOTE_REPO was not expanding $GITHUB_TOKEN and $GITHUB_REPOSITORY.

PR #158, opened 2019-11-27
toSwiftVariableCase() shouldn't be adding `` to variable names in the runtime.

adam-fowler:
It appears we have been adding `` around keywords when accessing their values using `Mirror`. This doesn't work. The code seems to be left over from the Code Generator where it is needed. This PR removes this. 

Added tests to verify this.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -421,7 +421,6 @@ extension AWSClient {
                     default:
                         body = Body(anyValue: payloadBody)
                     }
-                    headers.removeValue(forKey: payload.toSwiftVariableCase())
                 } else {
                     body = .empty
                 }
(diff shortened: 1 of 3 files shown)


adam-fowler:
The CodeGenerator used to use AWSSDKSwiftCore for some of its functionality. Including the function mentioned in this PR.  (I removed that dependency a while back)

Obviously we can’t trust amazon not to use swift keywords so we add speech marks around any variables that are swift keywords in the Code Generator. But we don’t need to do that in the runtime when using Mirror. If we do that it doesn’t actually work.

fabianfett:
That makes sense. As I mentioned, this is a sensible thing to change, and I understand and favor the goal. I just wanted to mention, that I am probably not the best guy to review this change, since I don't know to much about the code generation anyways.

adam-fowler:
I think the guy who wrote this stuff left the building at least a year ago. If you don’t want to put your name to it don’t worry. I’ll take the responsibility

fabianfett:
No risk, nof fun ;)

PR #159, opened 2019-12-02
Performance tests

adam-fowler:
Added tests to measure various sections of aws-sdk-swift-core. Tests are split into three sections, create aws request, sign request and validate response.

With these I found we were wasting time creating a date formatter every time we processed a timestamp. Also found the unnecessary adding of `` around keywords, which was really expensive and then also breaking requests with swift keywords #158

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSShapes/TimeStamp.swift
+++ b/Sources/AWSSDKSwiftCore/AWSShapes/TimeStamp.swift
@@ -43,7 +43,9 @@ public struct TimeStamp {
     }
     
     /// date formatter for outputting dates
-    private static var defaultDateFormatter : DateFormatter {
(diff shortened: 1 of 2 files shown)


Sources/AWSSDKSwiftCore/AWSShapes/TimeStamp.swift
```diff
@@ -43,7 +43,9 @@ public struct TimeStamp {
     }
     
     /// date formatter for outputting dates
-    private static var defaultDateFormatter : DateFormatter {
+    private static var defaultDateFormatter : DateFormatter = createDefaultDateFormatter()
```
fabianfett: I would change the `var` to `let` in this case. Otherwise this is unsafe. Especially considering threadSafety.

PR #160, opened 2019-12-05
Injectable CredentialProvider for AsyncHTTPClient approach

fabianfett:
This is the PR to the changes in https://github.com/adam-fowler/aws-signer/pull/2. Thanks @adam-fowler for preparing the ground already so far, that we can abstract away the process of getting credentials for signing. With this there are no more shared instances in the process (`MetaDataService.getCredential`).

Main focuses of this PR:

1. The current `AWSClient.init` has become a convenience initializer to ensure source compatibility. There is a new initializer where all properties are mandatory. The new one allows us to inject a couple of protocol based implementations `CredentialsProvider`, `EventLoopGroup`, `AWSHTTPClient`.

2. A `CredentialProvider` is a protocol abstracting away the process of getting the `AWSSigner.Credential`. Since returning the credential might require a network call, the function to get the credential is async: func getCredential() -> EventLoopFuture<Credential> as you can see I return a EventLoopFuture. A really simple implementation of this can be found in file `StaticCredentialProvider` in line 20.

3. There is a `StaticCredentialProvider`, a `EnvironmentCredentialProvider` and a `SharedCredentialProvider`

4. I implemented a `MetaDataCredentialProvider<MetaDataClient>` that is thread safe and only needs a `MetaDataClient` (another protocol to get the MetaData from ECS or EC2).

I think with this approach we can have a number of advantages:

1. The AWSClient can be initialized with a CredentialProvider. If none is handed over, we have a `CredentialsChain` (in file `CredentialProvider`), in which we select the best provider for the current use-case and complete lifecycle of the `AWSClient`. But this allows to use a non default profile in SharedCredential which is needed for using multiple accounts with one key. 🚀

2. We have a single `CredentialProvider` for the complete lifecycle of the `AWSClient`. We don't need a lock in the `AWSClient`. Locking the credential state is IMHO an implementation detail of the specific `CredentialProvider`. With this the lock is exactly there (see `MetaDataCredentialProvider`). StaticCredentials won't use a lock at any point.

3. Besides `ByteBuffer` AWSSigner is completely independent from NIO which probably could be even further abstracted away with a protocol if wanted. Therefore AWSSigner can be used in iOS projects that need to sign requests against APIGateways with custom domains.


adam-fowler:
So let me get this right. You've extracted the credentials from the signer and supply them every time you sign a request. You have created a EventLoopFuture<> that provides the credentials when needed. The lock has been moved to when we are updating the credentials future. It currently seems a quite a large amount of code changes to implement this. I'm sure we can do this with a smaller amount of change.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -284,72 +292,43 @@ extension AWSClient {
...
-    }
-
-    func createHTTPRequest(_ awsRequest: AWSRequest, signer: AWSSigner) -> AWSHTTPRequest {
-        // if credentials are empty don't sign request
-        if signer.credentials.isEmpty() {
-            return awsRequest.toHTTPRequest()
-        }
+    func signRequest(_ awsRequest: AWSRequest, with credential: Credential) -> AWSHTTPRequest {
+        
+        #warning("we need a way if we don't have credentials here. is that even possible?")
```
adam-fowler: You removed the ability to created unsigned requests. There are a number of AWS services that allow for unsigned requests (eg CognitoIdentityProvider)

adam-fowler:
From what I can work out all your changes could be reduced to 
```swift
protocol CredentialFuture {
    mutating func getCredential() -> Future<CredentialProvider>
}

struct StaticCredentialFuture: CredentialFuture {
    let credential: CredentialProvider
    let eventLoopGroup: EventLoopGroup

    init(credential: CredentialProvider,  eventLoopGroup: EventLoopGroup) {
        self.credential = credential
        self.eventLoopGroup = eventLoopGroup
    }

    func getCredential() -> Future<CredentialProvider> {
        return eventLoopGroup.next().makeSucceededFuture(credential)
    }
}

class MetaDataCredentialFuture: CredentialFuture {
    var credential: CredentialProvider
    let httpClient: AWSHTTPClient
    let lock = NSLock()

    init(httpClient: AWSHTTPClient) {
        self.httpClient = httpClient
        self.credential = ExpiringCredential(accessKeyId: "", secretAccessKey: "", expiration: Date.init(timeIntervalSince1970: 0))
    }
    
    func getCredential() -> Future<CredentialProvider> {
        if let expiringCredential = self.credential as? ExpiringCredential, expiringCredential.nearExpiration() {
            return refresh()
        }
        return httpClient.eventLoopGroup.next().makeSucceededFuture(self.credential)
    }
    
    func refresh() -> Future<CredentialProvider> {
        return MetaDataService.getCredential(httpClient: self.httpClient).map { credential in
            self.lock.lock()
            self.credential = credential
            self.lock.unlock()
            return credential
        }
    }
}
```
Once you have your credential you can then create an AWSSigner with that and sign your request. I'm not a bug fan of the splitting of AWSSigner into credential data and the code that uses that credential data.

adam-fowler:
I guess my solution would need the future setting/clearing in refresh() as well

fabianfett:
Hi Adam, I can fully understand that the proposed changes are an awful lot for the state the `AsyncHTTPBranch` is in.

There are a couple of things, that I really like about this approach, that maybe were missed in your review (at least you didn't bring them up).

Most of the issues I'm going to lay out were not introduced in the `AsyncHTTPClient` pr and I'm fully aware that the `AsyncHTTPClient` pr tries to introduce the `AsyncHTTPClient` with as few changes to the current code base as possible. Therefore I'm totally in off not going all in right now.

I try to make as many comments as possible in the `AsyncHTTPClient` branch. And well maybe I can rebase the bigger changes here for another PR in the future.

PR #162, opened 2019-12-05
Async http client credential provider

adam-fowler:
Simplified version of CredentialProvider that Fabian posted earlier
- Renamed Credential to StaticCredential
- Renamed CredentialProvider to Credential
- Added protocol CredentialProvider which returns a EventLoopFuture< Credential >
- Added StaticCredentialProvider which always returns the credential it was initialised with
- Added MetaDataCredentialProvider which returns credentials supplied by MetaDataService. It also manages the renewal of these.
- AWSClient is initialised with CredentialProvider
- send() functions calls CredentialProvider.getCredential() and in the flatMap of the result it creates an AWSSigner to sign our request.

Tested on EC2, with expiration dates being forced to 5 mins and 3 seconds, so we'll see a renew of credentials.


Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -56,7 +32,9 @@ public final class AWSClient {
         case useAWSClientShared
     }
 
-    var signer: AtomicProperty<AWSSigner>
+    var credentialProvider: CredentialProvider
```
fabianfett: I think this can be a `let`

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
```diff
@@ -0,0 +1,61 @@
...
+//
+//
+import AWSSigner
+import Foundation
+import NIO
+import NIOConcurrencyHelpers
+
+/// Protocol providing future holding a credential
+protocol CredentialProvider {
+    mutating func getCredential() -> EventLoopFuture<Credential>
```
fabianfett: I understand, that we have need the `mutating` aspect in order to allow `struct`s. I'm not a big fan though. Since the credentialProvider is just initialized once, the benefits of a `struct` vs a `class` are IMHO neglectable.

Sources/AWSSigner/signer.swift
```diff
@@ -12,18 +12,18 @@ import NIO
...
+public struct AWSSigner {
     /// security credentials for accessing AWS services
-    public let credentials: CredentialProvider
+    public var credentials: Credential
     /// service signing name. In general this is the same as the service name
-    public let name: String
+    public var name: String
     /// AWS region you are working in
-    public let region: String
+    public var region: String
```
fabianfett: I wouldn't make these properties mutable. What would the usecase be?

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    func getCredential() -> EventLoopFuture<Credential> {
+        return eventLoopGroup.next().makeSucceededFuture(credential)
+    }
+}
+
+/// Provides credentials acquired from the metadata service. These expire and need updated occasionally
+class MetaDataCredentialProvider: CredentialProvider {
+    var credentialFuture: EventLoopFuture<Credential>
```
fabianfett: this should be `private(set) var`

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
```diff
@@ -0,0 +1,61 @@
...
+    let httpClient: AWSHTTPClient
+    let lock = Lock()
+
+    init(httpClient: AWSHTTPClient) {
+        let credential = ExpiringCredential(accessKeyId: "", secretAccessKey: "", expiration: Date.init(timeIntervalSince1970: 0))
+        self.credentialFuture = httpClient.eventLoopGroup.next().makeSucceededFuture(credential)
```
fabianfett: I'm not a huge fan using the `refreshFuture` all the time. IMHO this future should only be used while refreshing. What is wrong with my proposed solution? It's longer. But in my biased view more readable.

```swift
    func getCredential() -> EventLoopFuture<Credential> {
        self.lock.lock()
        let cred = credential
        self.lock.unlock()
        
        if let cred = cred, cred.nearExpiration() == false {
            // we have credentials and those are still valid
            
            if self.eventLoopGroup is MultiThreadedEventLoopGroup {
              // if we are in a MultiThreadedEventLoopGroup we try to minimize hops.
              return MultiThreadedEventLoopGroup.currentEventLoop!.makeSucceededFuture(cred)
            }
            return self.eventLoopGroup.next().makeSucceededFuture(cred)
        }
        
        // we need to refresh the credentials
        return self.refreshCredentials()
    }
    
    private func refreshCredentials() -> EventLoopFuture<Credential> {
        self.lock.lock()
        defer { self.lock.unlock() }
        
        if let future = credentialFuture {
            // a refresh is already running
            return future
        }
        
        credentialFuture = self.metaDataClient.getMetaData()
            .map { (metadata) -> (Credential) in
                let credential = metadata.credential
                
                self.lock.lock()
                defer { self.lock.unlock() }
                
                self.credentialFuture = nil
                self.credential = credential
                
                return credential
            }

        return credentialFuture!
    }
```

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
```diff
@@ -0,0 +1,61 @@
...
+        self.httpClient = httpClient
+    }
+    
+    func getCredential() -> EventLoopFuture<Credential> {
+        return credentialFuture.flatMap { credential in
+            return self.lock.withLock {
+                if let expiringCredential = credential as? ExpiringCredential, expiringCredential.nearExpiration() {
+                    return self.refresh()
+                } else {
+                    return self.credentialFuture
```
fabianfett: returning the future, that we are currently flatmapping on. Not a fan.

Sources/AWSSDKSwiftCore/AWSClient.swift
adam-fowler: yep you're right

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
adam-fowler: that was left from me trying something else. removed

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
adam-fowler: added

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
adam-fowler: why?

Sources/AWSSigner/signer.swift
adam-fowler: reverted, again was trying something forgot to revert

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
adam-fowler: - I guess the one thing I didn't like was we were holding two references to the credential, one inside the future and one in the class.
- The other thing was the optional credential, we shouldnt need an optional value there
I will implement the other version and we can compare.

adam-fowler:
Commited a form of your version. Slight changes are
- I used the withLock() function you get with the NIO lock.
- I removed the force unwrap of currentEventLoop as it was causing a crash
- credential is never allowed to be nil

adam-fowler:
Given `signURL(url:, httpMethod:, expires:)` now returns a `EventLoopFuture<URL>` we have an interface change and will be required to up to version 5.0.0

PR #163, opened 2019-12-12
Hypertext application language process refactor

adam-fowler:
- Moved HAL code into separate file
- Removed link loading as the code doesn't work and never seems to get used
- Added test for processing json+hal files

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient+HAL.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient+HAL.swift
@@ -1,0 +1,39 @@
+import HypertextApplicationLanguage
+import Foundation
+
+extension AWSClient {
+    /// process hal+json date. Extract properties from HAL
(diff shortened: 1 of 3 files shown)


Sources/AWSSDKSwiftCore/AWSClient+HAL.swift
```diff
@@ -0,0 +1,39 @@
+import HypertextApplicationLanguage
+import Foundation
+
+extension AWSClient {
+    /// process hal+json date. Extract properties from HAL
+    func hypertextApplicationLanguageProcess(response: AWSResponse, members: [AWSShapeMember]) throws -> AWSResponse {
+        switch response.body {
```
fabianfett:
```
guard case let .json(let data) = response.body,
  let contentType = response.headers["Content-Type"] as? String,
  content.contains("hal+json")
  else 
{
  return response
}
```

with this we can save two steps of indentation in the main code.

adam-fowler: done

Yasumoto:
Belated thank you!

PR #177, opened 2020-01-19
SwiftNIO dependency minor version bump

fabianfett:
Since we use the new `NIOAtomic` syntax introduced with SwiftNIO 2.11.0 we need to minor version bump our dependency.

PR that introduced new syntax: https://github.com/apple/swift-nio/pull/1263

cc: @weissi is such a move okay?


weissi:
@fabianfett Yes, minor version bumps are okay. Major ones usually require a major bump too (unless you're 100% sure not to expose any types from the dependency in your API).

PR #179, opened 2020-01-19
Remove all `import Foundation` and replace with specific imports.

fabianfett:
This will make much clearer where we depend on `Foundation`. In the future we can start to get rid of those parts.

Doesn't need to be merged now. Maybe rebase on master and reevaluate after `AsyncHTTPClient` has landed?


adam-fowler:
Have you compiled this on Linux, was wondering about the ISO8601DateFormatter. 

Also noticed pull requests to branches other than master don't run tests

Yasumoto:
In general, big fan of being explicit with our imports 🙇‍♂️

fabianfett:
close in favor off: #183

PR #181, opened 2020-01-19
Paginated results

adam-fowler:
This is the core routine for paginating results. It uses `Mirror` to extract the results and a continuation token to get the remaining results. The code generation PR can be found here https://github.com/swift-aws/aws-sdk-swift/pull/227

At the moment to create a new request with the new continuation token I have had to create a protocol which requires a new initialiser which take the original request and a token. The initialiser should copy the contents of the original request but with the new token.

The code below demonstrates the extension for S3.ListObjectsV2 and the paginate function used to get the data.
```
extension S3.ListObjectsV2Request: AWSPaginateable {
    public init(_ original: S3.ListObjectsV2Request, token: String) {
        self.init(
            bucket: original.bucket, 
            continuationToken: token, 
            delimiter: original.delimiter, 
            encodingType: original.encodingType, 
            fetchOwner: original.fetchOwner, 
            maxKeys: original.maxKeys, 
            prefix: original.prefix, 
            requestPayer: original.requestPayer, 
            startAfter: original.startAfter
        )
    }
}

extension S3 {
    public func listObjectsV2Paginator(_ input: ListObjectsV2Request) -> EventLoopFuture<[Object]> {
        return client.paginate(input: input, command: listObjectsV2, resultKey: "contents", tokenKey: "nextContinuationToken")
    }
}
```

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient+Paginate.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient+Paginate.swift
@@ -9,8 +9,12 @@ import NIO
 
 /// protocol for all AWSShapes that can be paginated.
 /// Adds an initialiser that does a copy but inserts a new pagination token
-public protocol AWSPaginateable: AWSShape {
+public protocol AWSPaginateStringToken: AWSShape {
     init(_ original: Self, token: String)
+}
+
+public protocol AWSPaginateIntToken: AWSShape {
+    init(_ original: Self, token: Int)
 }
 
 public extension AWSClient {
@@ -23,7 +27,7 @@ public extension AWSClient {
     ///   - command: Command to be paginated
     ///   - resultKey: The name of the objects to be paginated in the response object
     ///   - tokenKey: The name of token in the response object to continue pagination
-    func paginate<Input: AWSPaginateable, Output: AWSShape, PaginateOutput>(input: Input, command: @escaping (Input)->EventLoopFuture<Output>, resultKey: String, tokenKey: String) -> EventLoopFuture<[PaginateOutput]> {
+    func paginate<Input: AWSPaginateStringToken, Output: AWSShape, PaginateOutput>(input: Input, command: @escaping (Input)->EventLoopFuture<Output>, resultKey: String, tokenKey: String) -> EventLoopFuture<[PaginateOutput]> {
         var list : [PaginateOutput] = []
         
         func paginatePart(input: Input) -> EventLoopFuture<[PaginateOutput]> {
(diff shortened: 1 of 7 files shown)


fabianfett:
This is a great start!

As far as I can see your proposed API encourages accumulation of all results before returning them to the user. This IMHO is something that is in contrast to how nio systems should work: Stream as much as you can. Keep memory pressure low.

Therefore I'd propose an API like the following:

```swift
struct PagingContext {
  let page: Int
  let eventLoop: EventLoop // making creation of futures easy
}

func paginate<Input: AWSPaginateStringToken, Output: AWSShape, Result>(
   input: Input, 
   command: @escaping (Input)->EventLoopFuture<Output>, resultKey: PartialKeyPath<Output>, tokenKey: KeyPath<Output, String?>) 
   onPage : @escaping ([Result], PagingContext) -> EventLoopFuture<BOOL>) 
   -> EventLoopFuture<Void> 
{
    // implementation
}
```

The onPage callback is executed on every page. The developer can decide, if they want to continue the paging by returning a true or false. We don't accumulate more data than one request. 

With this, a developer can create a request, can consume the response pages one by one, and always easily cancel the next request.

Building an array of all objects should be for the developer fairly easy, but this is not something that I would not encourage with our API. If a user want's to iterate all files in S3, then this is fine for us. But given that they might have bezillions of files in there, we should not be the ones who use all the memory, by aggregating the results. ;)

adam-fowler:
I can see where you are coming from. I still think most people will collate the results into one array. There are many situations where the memory load won't be an issue. It would be easy enough to write a callback to collate arrays. We could provide one if necessary (although I know you wouldnt be keen on that, but if someone is writing the same wrapper repeatedly...).

Regarding the paging context I'm not sure of the necessity to return a page index. I can see how the eventloop might be of use but what eventloop are we passing in `client.eventLoopGroup.next()`? Given eventloops aren't passed to NIO closures I'm not sure we need to do it here either. Also we could ensure the closure is running on the eventloop via `submit`.

Both page index and eventloop can be accessed through variable capture.

adam-fowler:
I'll rewrite with closure, we can discuss the PagingContext more

fabianfett:
> We could provide one if necessary (although I know you wouldnt be keen on that, but if someone is writing the same wrapper repeatedly...).

May I ask, what your use case is for that? 

> Regarding the paging context I'm not sure of the necessity to return a page index. 

Fair point, I just wanted to be a good fellow. ;)

> I can see how the eventloop might be of use but what eventloop are we passing in client.eventLoopGroup.next()? Given eventloops aren't passed to NIO closures I'm not sure we need to do it here either. 

We need the EventLoop, to be able to make a succeeded future right away, to continue paging:
```
return eventLoop.makeSucceededFuture(true)
```
Also we could ensure the closure is running on the eventloop via submit.

> That's IMHO not the issue

May I ask in which context you need the whole thing?

adam-fowler:
Regarding EventLoop I guess we are looking at either providing the equivalent of a `flatMap` call which is required to return a `EventLoopFuture<Bool>` or a map call which returns a `Bool`. I was thinking of providing a `map` call. You have suggested a `flatMap` call.

Why would I need the collated results of a paginated call? Maybe I want to order results after they have come in. Any process that involves the whole array instead of individual elements will need collated results.

adam-fowler:
I’ve implemented the flat map version.

fabianfett:
Sorry. My code wasn't changed at all the positions it should have. Updated it now. paginatePart(input: Input) does not return a Future anymore. So those aren't stacked into each other, and hopefully are memory released. But I guess this should be fairly easy to test.

Sources/AWSSDKSwiftCore/AWSClient+Paginate.swift
```diff
@@ -0,0 +1,116 @@
...
+/// protocol for all AWSShapes that can be paginated.
+/// Adds an initialiser that does a copy but inserts a new string based pagination token
+public protocol AWSPaginateStringToken: AWSShape {
+    init(_ original: Self, token: String)
+}
+
+/// protocol for all AWSShapes that can be paginated.
+/// Adds an initialiser that does a copy but inserts a new integer based pagination token
+public protocol AWSPaginateIntToken: AWSShape {
+    init(_ original: Self, token: Int)
```
fabianfett: I'm not a big fan of the naming in this case. I doesn't feel swifty: 

URL is an example we can have a look at...
Creating a new instance by changing one property: `appendingPathComponent(_:)`

This is why I propose:
```
func usingPaginationToken(_:) -> Self
```

https://developer.apple.com/documentation/foundation/url/1780239-appendingpathcomponent

Tests/AWSSDKSwiftCoreTests/TestServer.swift
```diff
@@ -0,0 +1,158 @@
...
+            // write outbound
+            let outputData: Data
+            switch serviceProtocol {
+            case .json:
+                outputData = try JSONEncoder().encode(result.output)
+            case .xml:
+                outputData = try XMLEncoder().encode(result.output).xmlString.data(using: .utf8) ?? Data()
+            }
+
+            var byteBuffer = byteBufferAllocator.buffer(capacity: 0)
```
fabianfett: capacity should be `data.count`

PR #183, opened 2020-01-23
Remove all `import Foundation` and replace with specific imports.

fabianfett:
NOTE: This can be merged into master!

This will make much clearer where we depend on `Foundation`. In the future we can start to get rid of those parts.

### DIFF
--- a/Sources/AWSSDKSwiftCore/String.swift
+++ b/Sources/AWSSDKSwiftCore/String.swift
@@ -6,8 +6,6 @@
 //
 //
 
-import Foundation
-
 extension String {
     public func lowerFirst() -> String {
         return String(self[startIndex]).lowercased() + self[index(after: startIndex)...]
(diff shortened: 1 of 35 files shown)


Sources/AWSSDKSwiftCore/String.swift
```diff
@@ -6,7 +6,42 @@
...
+    "nil",
+    "in",
+    "import",
+    "as",
+    "is",
+    "do",
+    "try",
+    "type",
+    "repeat"
+]
```
adam-fowler: and why is this here?

fabianfett: cherry pick. damn it!

fabianfett: fixed.

adam-fowler: I'm not a keen fan of cherry pick for this very reason

Issue #148, opened 2019-11-20
Support Instance Metadata v2


Yasumoto:
AWS [just announced](https://aws.amazon.com/about-aws/whats-new/2019/11/announcing-updates-amazon-ec2-instance-metadata-service/) there's a new, session-based Metadata Service API.

Implementation details [in the EC2 User Guide](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#configuring-instance-metadata-service).

adam-fowler:
Looks like we should be careful implementing this. Even AWS are getting this wrong. https://github.com/aws/aws-cli/issues/4682

adam-fowler:
done in #184

Issue #178, opened 2020-01-19
Refactor MetaDataService.getCredential(httpClient: self.httpClient)


fabianfett:
Once the `AsyncHTTPClient` has landed in master, I would really like to refactor the `MetaDataService` in so far that it doesn't use static methods (that involve some kind of decision making) anymore. `static func getCredential(httpClient: self.httpClient)` I'm looking at you.

This idea was originally proposed in #160 but left to not make the already huge changes even larger.

```swift
import Foundation
import NIO
import NIOConcurrencyHelpers
import AWSSigner

public struct ExpiringCredential: Credential {
  
    public let accessKeyId: String
    public let secretAccessKey: String
    public let sessionToken: String?
    public let expiration: Date?
    
    public init(accessKeyId: String, secretAccessKey: String, sessionToken: String? = nil, expiration: Date? = nil) {
        self.accessKeyId = accessKeyId
        self.secretAccessKey = secretAccessKey
        self.sessionToken = sessionToken ?? ProcessInfo.processInfo.environment["AWS_SESSION_TOKEN"]
        self.expiration = expiration
    }
    
    func nearExpiration() -> Bool {
        guard let expiration = self.expiration else {
            return false
        }
      
        // are we within 5 minutes of expiration?
        return Date().addingTimeInterval(5.0 * 60.0) > expiration
    }
}

/// protocol for decodable objects containing credential information
public protocol CredentialContainer: Decodable {
    var credential: ExpiringCredential { get }
}

public protocol MetaDataClient {
    associatedtype MetaData: CredentialContainer & Decodable
    
    func getMetaData() -> EventLoopFuture<MetaData>
}

enum MetaDataClientError: Error {
    case failedToDecode
    case couldNotGetInstanceRoleName
}

extension MetaDataClient {
    
    /// decode response return by metadata service
    func decodeResponse(_ bytes: ByteBuffer) throws -> MetaData {
        let decoder = JSONDecoder()
        // set JSON decoding strategy for dates
        let dateFormatter = DateFormatter()
        dateFormatter.locale = Locale(identifier: "en_US_POSIX")
        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss'Z'"
        dateFormatter.timeZone = TimeZone(secondsFromGMT: 0)
        decoder.dateDecodingStrategy = .formatted(dateFormatter)
        // decode to associated type
        let metaData = try decoder.decode(MetaData.self, from: bytes)
[...]

PR #184, opened 2020-01-23
Instance metadata service v2

adam-fowler:
Support new AWS instance metadata scheme that requires a session key. More details can be found here #148.

Tested against EC2 instance which requires IMDSv2 and also one that still also allows IMDSv1.

Also added a test to fully test the metadata service. To avoid issues with test failing in CI this test only runs if have the environment variable TEST_EC2_METADATA set.


Sources/AWSSDKSwiftCore/MetaDataService.swift
```diff
@@ -23,7 +23,9 @@ import class  Foundation.ProcessInfo
 /// errors returned by metadata service
 enum MetaDataServiceError: Error {
     case missingRequiredParam(String)
+    case couldNotGetInstanceSessionKey
```
fabianfett: Currently not used

Sources/AWSSDKSwiftCore/MetaDataService.swift
```diff
@@ -180,32 +184,48 @@ struct InstanceMetaDataServiceProvider: MetaDataServiceProvider {
...
         // instance service expects absoluteString as uri...
-        return request(uri:InstanceMetaDataServiceProvider.baseURLString, timeout: 2, eventLoopGroup: eventLoopGroup)
-            .flatMapThrowing{ response in
-                switch response.head.status {
-                case .ok:
-                    let roleName = String(data: response.body, encoding: .utf8) ?? ""
-                    return "\(InstanceMetaDataServiceProvider.baseURLString)/\(roleName)"
-                default:
-                    throw MetaDataServiceError.couldNotGetInstanceRoleName
+        return request(uri:InstanceMetaDataServiceProvider.apiTokenURL, method: .PUT, headers:["X-aws-ec2-metadata-token-ttl-seconds":"21600"], timeout: 2, eventLoopGroup: eventLoopGroup).flatMapThrowing { response in
```
fabianfett: line breaks? maybe? please ;)

what do you think about return an optional dict of headers?

Sources/AWSSDKSwiftCore/MetaDataService.swift
```diff
@@ -180,32 +184,48 @@ struct InstanceMetaDataServiceProvider: MetaDataServiceProvider {
...
-            .flatMap { uri in
-                return self.request(uri: uri, timeout: 2, eventLoopGroup: eventLoopGroup)
-            }
-            .map { response in
+                return
+            }.flatMapError { error in
+                // If we didn't find a session key then assume we are running IMDSv1 (we could be running from a Docker container
+                // and the hop count for the PUT request is still set to 1)
+                print(error)
+                return eventLoopGroup.next().makeSucceededFuture(())
```
fabianfett: flatMapError to optional headers -> means we don't have headers here -> return nil

Sources/AWSSDKSwiftCore/MetaDataService.swift
```diff
@@ -180,32 +184,48 @@ struct InstanceMetaDataServiceProvider: MetaDataServiceProvider {
...
+            }.flatMap { _ in
```
fabianfett:
```swift
  }.flatMap { header in
    let header = header ?? [:]
    // request rolename
    return self.request(uri:InstanceMetaDataServiceProvider.baseURLString, headers: header, timeout: 2, eventLoopGroup: eventLoopGroup)
```

Sources/AWSSDKSwiftCore/MetaDataService.swift
```diff
@@ -180,32 +184,48 @@ struct InstanceMetaDataServiceProvider: MetaDataServiceProvider {
...
+                // request rolename
+                return self.request(uri:InstanceMetaDataServiceProvider.baseURLString, headers:sessionTokenHeader, timeout: 2, eventLoopGroup: eventLoopGroup)
+            }.flatMapThrowing { response in
+                // extract rolename
+                guard response.head.status == .ok,
+                    let roleName = String(data: response.body, encoding: .utf8) else { throw MetaDataServiceError.couldNotGetInstanceRoleName }
```
fabianfett: line break? maybe? please ;)

Tests/AWSSDKSwiftCoreTests/MetaDataServiceTests.swift
```diff
@@ -7,17 +7,32 @@
...
-  func testMetaDataServiceForECSCredentials() {
+    // Disabling cannot guarantee that 169.254.169.254 is not a valid IP on another network
+    func testMetaDataGetCredential() {
+        if ProcessInfo.processInfo.environment["TEST_EC2_METADATA"] != nil {
+            do {
+                let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+                _ = try MetaDataService.getCredential(eventLoopGroup: group).wait()
+            } catch {
+                print(error)
+                XCTFail()
```
fabianfett: instead of print, use 
```
XCTFail("Unexpected error: \(error)")
```

This way the error is shown directly in the xcode red bubble next to the failing error. No search in console required.

fabianfett:
Mostly LGTM, small nitpicks, see review. I guess, we can improve the metadatas testing as soon as #178 is implemented.

Sources/AWSSDKSwiftCore/MetaDataService.swift
adam-fowler: you caught me, mid development here. Thought I was done yesterday but realised this morning there was a issue with running containers on EC2 and this setup

Sources/AWSSDKSwiftCore/MetaDataService.swift
adam-fowler: The sessionTokenHeaders are used in multiple places (I also need them when getting credentials) within the process. I could pass them down through all the flatMaps but thats quite awkward so just leaving them at the top and using variable capture seemed easier

Sources/AWSSDKSwiftCore/MetaDataService.swift
adam-fowler: ok

Tests/AWSSDKSwiftCoreTests/MetaDataServiceTests.swift
adam-fowler: This test nevers run in XCode. It is only ever of use on an EC2 instance. I'll change it anyway

Issue #172, opened 2020-01-18
Make request cancellable


fabianfett:
Use the same interface that AysncHTTPClient uses:

```swift
protocol HTTPClient {
  func execute(request: AWSHTTPRequest, timeout: TimeAmount) -> AWSTask< AWSHTTPResponse >
}

class AWSTask<Response: AWSHTTPResponse > {
  let response: EventLoopFuture<Response>
  let eventLoop: EventLoop { get }

  func cancel()
}
```

fabianfett:
Yes! If a web request is dropped (in a server use-case), there should be ways to stop the work done (whatever it is).

adam-fowler:
What does cancel actually mean? Can we cancel a request once it has been sent.

fabianfett:
Well the same it means to cancel an HTTP Request. Just drop the connection.

Issue #165, opened 2019-12-17
Initialise AWS client with HTTPClient


adam-fowler:
Allow user to pass in HTTPClient to AWSClient.init() and use that for all http calls. This would allow the user to pass in an HTTPClient configured as they like. Perhaps they have a proxy to configure.

If we use the AWSHTTPClient protocol defined in the async-http-client branch they could use their own client as long as it conforms to the AWSHTTPClient.

This does beg the question do we need to pass in an eventLoopGroupProvider if we allow the user to provide the HTTP client as the eventLoopGroup is only ever used in conjunction with the HTTPClient or to provide succeeded/failed futures? I'm guessing we could support both but providing both at the same time would be an invalid operation.

fabianfett:
I would love that! Probably something for 5.0.

fabianfett:
> This does beg the question do we need to pass in an eventLoopGroupProvider if we allow the user to provide the HTTP client as the eventLoopGroup is only ever used in conjunction with the HTTPClient or to provide succeeded/failed futures? I'm guessing we could support both but providing both at the same time would be an invalid operation.

I'm not sure if we should compromise on this point: I came to `aws-sdk-swift` because it is undoubtedly a `swift-nio` based framework. I can see the benefits of giving users more flexibility but I'm afraid this would be at the costs of its server side focus.

adam-fowler:
> I'm not sure if we should compromise on this point: I came to `aws-sdk-swift` because it is undoubtedly a `swift-nio` based framework. I can see the benefits of giving users more flexibility but I'm afraid this would be at the costs of its server side focus.

The reason I mention not passing in an eventLoopGroupProvider is the only situation we use the eventLoopGroup is as part of the HttpClient. If we are supplying the HttpClient it can setup it's own eventLoopGroup. And the one we pass into AWSClient isn't used for anything.

fabianfett:
Oh true. My bad. That is very true. In favor!

fabianfett:
Question would be than though how would we implement an `AWSHTTPClient` based on `URLSession` that emits `EventLoopFuture`s. To be honest this sounds more like an intellectual challenge than something at least I would use.

PR #186, opened 2020-01-28
Allow users to specify custom timeouts for requests

mrabiciu:
On a slow or busy network 5 seconds is sometimes not enough. 
This diff adds a custom timeout parameter to all the send methods and defaults to the previous 5 second timeout.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -154,9 +154,9 @@ public class AWSClient {
 }
 // invoker
 extension AWSClient {
-    fileprivate func invoke(_ nioRequest: HTTPClient.Request) -> EventLoopFuture<HTTPClient.Response> {
+    fileprivate func invoke(_ nioRequest: HTTPClient.Request, timeout: TimeAmount = .seconds(5)) -> EventLoopFuture<HTTPClient.Response> {
         let client = HTTPClient(eventLoopGroupProvider: .shared(eventLoopGroup))
-        let futureResponse = client.connect(nioRequest)
+        let futureResponse = client.connect(nioRequest, timeout: timeout)
 
         futureResponse.whenComplete { _ in
             do {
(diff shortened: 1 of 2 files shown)


adam-fowler:
We are looking to allow the user to provide their own HTTPClient in the near future #165. You would be able to control the timeout's from there once that work is in place. So I'm not sure this is necessary.

fabianfett:
Hi @mrabiciu, thanks so much for your contribution. It's always exciting to see new developers getting involved with this project!

I think that your intent is certainly of relevance and that we should do something in this area. 

As @adam-fowler has pointed out our current thinking focuses (we are currently in the planning/ beginning of version 5.0) on giving the user the option to create the [AsyncHTTPClient](https://github.com/swift-server/async-http-client) themselves and injecting the client into the AWSClient. (See #165) The AWSClient itself could then be used to initialise the service facade (See swift-aws/aws-sdk-swift#231). With this approach you could configure the timeout interval on the AsyncHTTPClient and the value would be valid for all calls.

While your approach would introduce some "clutter" to the API, I can also see the point that changing the timeout interval could be necessary on a per call basis. Using another AWSClient with a separate AsyncHTTPClient can not be in our interest, since it would limit connection pooling. 

In the end I think, you are right and we should accept your change (I think i'm in favor, for the same reasons I proposed #172).

Would you be okay, if we put it on the list for 5.0? While it is not a breaking API change, I'd be in favor if we roll this out with a major release.

Does this sound okay to you?

adam-fowler:
Hi @mrabiciu yes thanks for your PR. 

@fabianfett I'm not convinced yet of the need for this. Is there a valid use case for setting the timeout on a call by call basis? This would be more clutter and we already have other PR/Issues suggesting additional parameters to the send() calls https://github.com/swift-aws/aws-sdk-swift/issues/231.

Also this PR is only the change to aws-sdk-swift-core. The timeout parameter would have to be passed to all the api calls in aws-sdk-swift. So every call would look like
```
let future = try s3.putObject(request, timeout: .seconds(20))
```

Another possibility is we store the timeout in the `ServiceConfig` from https://github.com/swift-aws/aws-sdk-swift/issues/231. 

Lets wait until the HTTPClient provider  and possibly the ServiceConfig work is complete and re-visit this then. 

@mrabiciu what are your thoughts?

adam-fowler:
@mrabiciu #328

adam-fowler:
This is now possible using by creating a copy of the service with a new time out value using `AWSService.timingOut(after:)`. Code would work as follows 
```
s3.timingOut(after: .seconds(60)).getObject(bucket: "mybucket", key: "mykey")
```
therefore I am closing this

PR #188, opened 2020-01-30
Fixed Dictionary<enum, object> in XMLEncoder/XMLDecoder plus optimization

adam-fowler:
Found this while implementing the QueryEncoder.

The container coding was not being reset in the UKEC encode/decode encode<T>(_) functions so the container coding was being applied to the child elements of the array. This only caused an issue when encoding/decoding dictionaries where the key was not a base type eg an enum and the value was a Codable type. The container coding for the UKEC was being passed to the KEC that coded the Codable type.


adam-fowler:
Added a minor optimization to avoid the need to store the current container encoding. I process the container encoding in the UKEC and KEC init() functions so don't need to store it during processing of the container.

PR #189, opened 2020-01-30
Remove HypertextApplicationLanguage framework

adam-fowler:
We aren't interested in the majority of hal+json output. The only things we need are the resources. Resources can be included as normal, or they can be included inside an "_embedded" tag. Instead of running the HAL code I just look for the "_embedded" tag and if it exists merge everything underneath with the top level json dictionary and output that.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient+HAL.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient+HAL.swift
@@ -1,36 +1,29 @@
-import HypertextApplicationLanguage
 import Foundation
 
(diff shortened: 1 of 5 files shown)


Sources/AWSSDKSwiftCore/AWSClient+HAL.swift
```diff
@@ -1,36 +1,29 @@
-import HypertextApplicationLanguage
 import Foundation
 
+// AWS HAL services I know of are APIGateway, Pinpoint, Greengrass
```
fabianfett: Awesome comment. Thanks so much!

PR #192, opened 2020-02-03
Use Apple swift-crypto

adam-fowler:
Use new swift-crypto library from Apple. This removes the requirement for an OpenSSL library. Woohoo! 

But it does come with some new minimum requirements. 
Swift 5.1
macOS 10.15
iOS 13


Sources/AWSSignerV4/signer.swift
```diff
@@ -141,28 +142,28 @@ public struct AWSSigner {
...
     /// Stage 2 Create the string to sign as in https://docs.aws.amazon.com/general/latest/gr/sigv4-create-string-to-sign.html
-    func stringToSign(signingData: SigningData) -> String {
+    func stringToSign(signingData: SigningData) -> Data {
         let stringToSign = "AWS4-HMAC-SHA256\n" +
             "\(signingData.datetime)\n" +
             "\(signingData.date)/\(region)/\(name)/aws4_request\n" +
-            AWSSigner.hexEncoded(sha256(canonicalRequest(signingData: signingData)))
-        return stringToSign
+            SHA256.hash(data: canonicalRequest(signingData: signingData)).hexDigest()
+        return Data(stringToSign.utf8)
```
fabianfett: We don't need to use Data here, do we? `stringToSign.utf8` should be an [UInt8] anyway.

Sources/AWSSignerV4/signer.swift
```diff
@@ -172,38 +173,36 @@ public struct AWSSigner {
...
+
     /// Create a SHA256 hash of the Requests body
     static func hashedPayload(_ payload: BodyData?) -> String {
         guard let payload = payload else { return hashedEmptyBody }
-        let hash : [UInt8]?
+        let hash : String?
         switch payload {
         case .string(let string):
-            hash = sha256(string)
+            hash = SHA256.hash(data: Data(string.utf8)).hexDigest()
```
fabianfett: Replace `Data(string.utf8)` with `[UInt8](string.utf8)`. This SHOULD be a non copy operation and we lower the dependency on `Data`.

Sources/AWSSignerV4/signer.swift
```diff
@@ -19,23 +20,23 @@ public struct AWSSigner {
...
     /// Initialise the Signer class with AWS credentials
     public init(credentials: Credential, name: String, region: String) {
         self.credentials = credentials
         self.name = name
         self.region = region
     }
-    
+
     /// Enum for holding your body data
     public enum BodyData {
```
fabianfett: While we are at it, I would love to see this removed and be replaced with a protocol. I even think that 
```
<Bytes: Collection> where Bytes.Element == UInt8
```

should be enough.

Sources/AWSSignerV4/signer.swift
```diff
@@ -172,38 +173,36 @@ public struct AWSSigner {
             "\(canonicalHeaders)\n\n" +
             "\(signingData.signedHeaders)\n" +
             signingData.hashedPayload
-        return canonicalRequest
+        return Data(canonicalRequest.utf8)
     }
-    
+
     /// Create a SHA256 hash of the Requests body
     static func hashedPayload(_ payload: BodyData?) -> String {
```
fabianfett: Replace with 

```swift
@inlineable static func hashedPayload<Bytes: Collection>(_ payload: Bytes?)  -> String where Bytes.Element == UInt8 
```

Sources/AWSSignerV4/signer.swift
fabianfett: I put down an example of the new method signature. I think this would make the signer code way better, and would lead to a signer, that maybe is only depended on `Crypto`. And we could make this dependency even

```swift
@_implementationOnly import Crypto
```

adam-fowler: > While we are at it, I would love to see this removed and be replaced with a protocol. I even think that
> 
> ```
> <Bytes: Collection> where Bytes.Element == UInt8
> ```
> 
> should be enough.

Isn't that simple. As the body is optional there are situations where it cannot ascertain what object `Bytes` is. Would need to have separate functions, one that takes a non optional buffer and one that does not. It is doable just more work.

adam-fowler: I've add another PR for this so you can see this done as separate functions #195

adam-fowler: Thought I tested it, obviously not. It fails for `Data` as `withContiguousStorageIfAvailable` doesn't work for `Data`. There is no other way to get a raw pointer with `Collection`. Don't think we can generalize this for `Collections`

Yasumoto:
Wow, so stoked we get to clean this all up!!

adam-fowler:
Closed in favour of #197

PR #195, opened 2020-02-06
Generic versions of sign functions where Body comforms to Collection

adam-fowler:
This forces us to provide two different implementations, one with a body and one without.


adam-fowler:
haha doesn't even work

adam-fowler:
I forgot withContiguousStorageIfAvailable{} which we need to get a pointer for the SHA256 function doesn't work for Data. There are no unsafe calls available for Collection. This is most likely a deadend

PR #197, opened 2020-02-07
AWS Swift Crypto

adam-fowler:
swift-crypto for Linux
CommonCrypto for macOS, iOS.

On Linux AWSCrypto just exports Crypto.
On AppleOS's it creates an API that replicates CryptoKit while using CommonCrypto. This removes the requirement for macOS 10.15 and iOS 13.

This is not ideal but I cannot see any other way of support older versions of macOS, iOS.


Sources/AWSCrypto/Digest.swift
```diff
@@ -0,0 +1,16 @@
+// Digest.swift
+// Replicating the CryptoKit framework interface for < macOS 10.15
+// written by AdamFowler 2020/01/30
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
fabianfett: replace with `#if canImport(CommonCrypto)`?

Sources/AWSCrypto/HMAC.swift
```diff
@@ -0,0 +1,68 @@
+// HMAC.swift
+// Replicating the CryptoKit framework interface for < macOS 10.15
+// written by AdamFowler 2020/01/30
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
fabianfett: replace with `#if canImport(CommonCrypto)`?

Package.swift
```diff
@@ -20,29 +19,31 @@ let package = Package(
...
+
+        .testTarget(name: "AWSCryptoTests", dependencies: ["AWSCrypto"]),
         .testTarget(name: "AWSSDKSwiftCoreTests", dependencies: ["AWSSDKSwiftCore", "NIOTestUtils"]),
         .testTarget(name: "AWSSignerTests", dependencies: ["AWSSignerV4"])
     ]
 )
 
-// switch for whether to use CAWSSDKOpenSSL to shim between OpenSSL versions
+// switch for whether to use swift crypto. Swift crypto requires macOS10.15 or iOS13.I'd rather not pass this requirement on
 #if os(Linux)
```
fabianfett: #if canImport(CommonCrypto)
let useSwiftCrypto = false
#else 
let useSwiftCrypto = true
#endif

Sources/AWSCrypto/HashFunction.swift
```diff
@@ -0,0 +1,71 @@
+// HashFunction.swift
+// HashFunction protocol and OpenSSL, CommonCrypto specialisations
+// Replicating the CryptoKit framework interface for < macOS 10.15
+// written by AdamFowler 2020/01/30
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
fabianfett: replace with `#if canImport(CommonCrypto)`?

Sources/AWSCrypto/Insecure.swift
```diff
@@ -0,0 +1,9 @@
+// Insecure.swift
+// Replicating the CryptoKit framework interface for < macOS 10.15
+// written by AdamFowler 2020/01/30
+
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
fabianfett: replace with `#if canImport(CommonCrypto)`?

Sources/AWSCrypto/MD5.swift
```diff
@@ -0,0 +1,43 @@
+// MD5.swift
+// Replicating the CryptoKit framework interface for < macOS 10.15
+// written by AdamFowler 2020/01/30
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
fabianfett: replace with `#if canImport(CommonCrypto)`?

Sources/AWSCrypto/SHA2.swift
```diff
@@ -0,0 +1,104 @@
+// SHA2.swift
+// Replicating the CryptoKit framework interface for < macOS 10.15
+// written by AdamFowler 2020/01/30
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
fabianfett: replace with `#if canImport(CommonCrypto)`?

Sources/AWSCrypto/SymmetricKey.swift
```diff
@@ -0,0 +1,28 @@
+// SymmetricKey.swift
+// Replicating the CryptoKit framework interface for < macOS 10.15
+// written by AdamFowler 2020/01/30
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
fabianfett: replace with `#if canImport(CommonCrypto)`?

Sources/AWSCrypto/exports.swift
```diff
@@ -0,0 +1,6 @@
+// Digest.swift
+// Replicating the CryptoKit framework interface for < macOS 10.15
+// written by AdamFowler 2020/01/30
+#if !(os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
```
fabianfett: replace with `#if !canImport(CommonCrypto)`?

docker/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM swift:5.1.3 as builder
```
fabianfett: Maybe we should update this one? `5.1.5`?

adam-fowler: Is 5.1.5 out? I thought we were at 5.1.4

Sources/AWSCrypto/Digest.swift
adam-fowler: This doesn't exactly sync with the test in Package.swift `#if os(Linux)`.

Package.swift
adam-fowler: But that isn't what we are testing for. We are testing for a platform. Maybe we should sync the platform tests instead

docker/Dockerfile
fabianfett: 5.1.5 is out:
https://swift.org/download/

Sources/AWSCrypto/Digest.swift
adam-fowler: `#if !os(Linux)` instead. See comment below

Package.swift
adam-fowler: It is a bit of a moot point in that they all mean the same thing

fabianfett: I don't know: The point is implicit if we check for the operating system and is explicit if we check for the availability of the framework. For this reason I'm pro framework check.

adam-fowler: Yeah but CommonCrypto is the fallback. I'm not choosing CommonCrypto because it is available. I am choosing it because I don't want to enforce os version requirements on iOS and macOS. The os is the explicit choice, the framework is the implicit choice.

docker/Dockerfile
adam-fowler: I set 5.1.5 on master branch. Didn't want to hide change in another PR

Package.swift
fabianfett: fair point. Then let's work with if `os(Linux)` here. Let's maybe even make checks around macOS versions and replace everywhere else with 

```swift
#if !canImport(Crypto)
```

because this way we make sure it is really only used if `Crypto` is not available.

adam-fowler: I'm good with that

adam-fowler: done

PR #199, opened 2020-02-11
Moved `toHTTPRequest()` from AWSClient to AWSRequest.

fabianfett:
### Motivation

The AWSClient encapsulates *a lot* of functionality. For this reason it is in the interest of the maintainers to distribute functionality to the objects it belongs to.

### Modifications

I moved the conceptionally static method `createHTTPRequest` from the AWSClient to AWSRequest where it is a great fit as an instance method.

### Result

Less Code on AWSClient. Clearer encapsulation of functionality on AWSRequest.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -205,7 +205,7 @@ extension AWSClient {
                         path: path,
                         httpMethod: httpMethod,
                         input: input)
-            return self.createHTTPRequest(awsRequest, signer: signer)
+            return awsRequest.createHTTPRequest(signer: signer)
         }.flatMap { request in
             return self.invoke(request)
         }.flatMapThrowing { response in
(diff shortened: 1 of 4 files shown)


fabianfett:
In a following PR I would like to move 

```swift
func createAWSRequest(operation operationName: String, path: String, httpMethod: String) throws -> AWSRequest
```
and
```swift
func createAWSRequest<Input: AWSShape>(operation operationName: String, path: String, httpMethod: String, input: Input) throws -> AWSRequest
```

 to AWSRequest as well using convenience initialisers.

adam-fowler:
> In a following PR I would like to move
> 
> ```swift
> func createAWSRequest(operation operationName: String, path: String, httpMethod: String) throws -> AWSRequest
> ```
> 
> and
> 
> ```swift
> func createAWSRequest<Input: AWSShape>(operation operationName: String, path: String, httpMethod: String, input: Input) throws -> AWSRequest
> ```
> 
> to AWSRequest as well using convenience initialisers.

That is not so obvious as to do that you would need to pass the AWSClient to the AWSRequest as the functions need config data from AWSClient

Issue #185, opened 2020-01-28
Add Payload object for Data blobs


adam-fowler:
Add a `Payload` shape to support both `Foundation.Data` and `NIO.ByteBuffer` for data storage in AWSShapes. Use this for any Foundation.Data that is also flagged as the payload for a request or response. eg `S3.PutObject`. This would allow us to pass data from one command response to another without having to convert to `Foundation.Data` and back to `NIO.ByteBuffer`, while still allowing people to use `Data.Foundation`.
```
enum Payload {
    case .data(Data)
    case .byteBuffer(ByteBuffer)
    case .empty
}
```
The main thing that complicates implementing this is ByteBuffer does not comply with Codable. We would have to jump through some hoops to get response decoding working correctly to insert the payload into the response struct. Request encoding should be ok as we shouldn't need to encode the payload as we are outputting a raw data body.

adam-fowler:
We could implement this for Request shapes without doing the Response shapes initially.

adam-fowler:
This came up when I was looking at using NIONonBlockingFileIO which returns a ByteBuffer, which at the moment I would have to convert to Data before passing to aws-sdk-swift.

adam-fowler:
We could extend this to include streamed data. 
```
enum Payload {
    case stream(()->EventLoopFuture<ByteBuffer>)
    ...
}
```

bridger:
Channels can also send data from a file directly to a socket without loading the file into memory. The type that supports this is `FileRegion`. It would be awesome to be able to upload using that.

adam-fowler:
> Channels can also send data from a file directly to a socket without loading the file into memory. The type that supports this is `FileRegion`. It would be awesome to be able to upload using that.

I’ll have a look into this but we would need AsyncHTTPClient to support this, which as of this point in time it doesn’t.

PR #200, opened 2020-02-11
Body.buffer holds a ByteBuffer instead of Data

adam-fowler:
The idea behind this is to keep data in ByteBuffer format for as long as possible. This should reduce the chance of converting between formats and copying data as we process buffers during the setup of a request or the processing of a response.
- Removed `Body.asData()` to stop converting back to `Data` 
- Removed `Body.asDictionary()`. It was a special case for JSON. 
- Error handling now uses `Codable` to decode errors as I removed `Body.asDictionary()`
- Added tests for a raw data payloads
- Changed MD5 calculation to use UnsafeBufferPointer so I can use it with ByteBuffers (although this change will be lost when we merge crypto in)
- Replace `Body.buffer(Data)` with `Body.buffer(ByteBuffer)`
- Removed `Body.stream`. It is not implemented. If we implement would probably do it differently

At the moment I have left `Body.json` holding a `Data` object as I was concerned about json being converted back and forward. This may not be an issue, will look into later on.


Sources/AWSSDKSwiftCore/Message/Body.swift
```diff
@@ -31,47 +29,34 @@ extension Body {
     /// initialize Body with Any. If it is Data, create .buffer() otherwise create a String describing the value
     init(anyValue: Any) {
```
fabianfett: Why do we have an initializer with `Any`? What are the usecases? That looks a little spooky to me.

adam-fowler: Yeah, I kinda stumbled on that one as well. The situation is we have a payload which is an arbituary object, could be a string, could be data. It is accessed through `Mirror`. The next stage of this work (the Payload object) may mean we can remove this.

https://github.com/swift-aws/aws-sdk-swift-core/blob/0dc610d31f25b0252bb5cbdfb8b76eb3cdfdd7b1/Sources/AWSSDKSwiftCore/AWSClient.swift#L383

fabianfett: Okay, do we want to open an issue for this?

adam-fowler: This is the work I was talking about https://github.com/swift-aws/aws-sdk-swift-core/issues/185

adam-fowler: Basically all non AWSShape payloads would be managed by this Payload enum. It is currently an enum but we could convert it to a wrapper for ByteBuffer but I had an idea we could also use it to inject streamed payloads as well.

adam-fowler: Added #211 as #185 doesn't detail cleaning up of this init() function

adam-fowler: You happy for this to go in now?

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -644,63 +647,89 @@ extension AWSClient {
...
+            guard case .json(let data) = response.body else { break }
+            errorMessage = try? JSONDecoder().decode(JSONError.self, from: data)
+            if errorMessage?.code == nil {
+                errorMessage?.code = response.headers["x-amzn-ErrorType"] as? String
+            }
 
         case .json:
-            code = bodyDict["__type"] as? String
-            message = bodyDict.filter({ $0.key.lowercased() == "message" }).first?.value as? String
+            guard case .json(let data) = response.body else { break }
```
fabianfett: I'm not sure, if we should handle the error here with the default `unknown error` further down. maybe protocol type mismatch error?

adam-fowler: The error handling I’ve pretty much inherited. I’ve found the error format the json services return is not totally consistent. I need to investigate this further. I’ll add an issue to look into this later. Would that be enough for now? This PR doesn’t change how aws-sdk-swift responds to errors it just makes the handling code a little cleaner.

PR #201, opened 2020-02-11
Created the ServiceConfig object

fabianfett:
I tried my very best to keep this PR as small as possible. The AWSClient basically only uses the `ServiceConfig` object for storage of its own properties.

This is created as a `Draft pull request` since I'm sure there is quite some discussion upon us whether `ServiceConfig` should be a class or a struct: 

I think this is a value type that just happens to be copied around a lot and not changed much. But it is a value type. For this reason I think I'd be hugely in favour of going the CoW way.

### DIFF
--- a/Sources/AWSSDKSwiftCore/ServiceConfig.swift
+++ b/Sources/AWSSDKSwiftCore/ServiceConfig.swift
@@ -1,0 +1,67 @@
+//
+//  ServiceConfig.swift
+//  AWSSDKSwiftCore
+//
+//  Created by Fabian Fett on 11.02.20.
+//
+
+import class Foundation.ProcessInfo
+
+public struct ServiceConfig {
+    public let region            : Region
+    public let amzTarget         : String?
+    public let service           : String
+    public let signingName       : String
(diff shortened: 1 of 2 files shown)


fabianfett:
For reference: https://github.com/swift-aws/aws-sdk-swift/issues/231

Sources/AWSSDKSwiftCore/ServiceConfig.swift
```diff
@@ -0,0 +1,67 @@
...
+    public let amzTarget         : String?
+    public let service           : String
+    public let signingName       : String
+    public let serviceProtocol   : ServiceProtocol
+    public let apiVersion        : String
+    public let endpoint          : String
+    public let serviceEndpoints  : [String: String]
+    public let partitionEndpoint : String?
+    public let possibleErrorTypes: [AWSErrorType.Type]
+
```
adam-fowler: You'll need to add middlewares here as well. Services can have specific middleware. We'll probably need middleware arrays on both client and service.

adam-fowler:
I have submitted #239 which covers all of this. This can probably be closed

PR #202, opened 2020-02-11
Adding support for retrying requests that fail

adam-fowler:
Add support for retrying requests when they fail. Give users complete control over how the client responds to a failed request.
- Added RetryController protocol with one function which returns a retryWaitTime or nil depending on whether the failed request should retry.
- Created NoRetry, ExponentialRetry and JitterRetry versions of RetryController
- Removed status code error processing from validate functions
- If status code is not 2xx then throw error
- Catch any errors from HTTPclient execute including errors from non 2xx status
- Send error to retryController along with attempt number.
- Retry HTTP request if it returns a retry wait time.

- Added tests for retrying after errors
- Extended AWSTestServer to have more control over when errors are thrown.


adam-fowler:
Question: what errors should we retry on?
HTTP errors 5xx and 429 are the most obvious errors to retry on.
I haven't added any connection failures yet.

adam-fowler:
Merging with the code disabled, by setting the `RetryController` to `NoRetry()`. Will create a separate PR for enabling this.

PR #203, opened 2020-02-11
Allow user to set the HTTPClient for AWSClient

adam-fowler:
Equivalent aws-sdk-swift PR can be found here https://github.com/swift-aws/aws-sdk-swift/pull/250

Added `httpClientProvider` parameter to `AWSClient.init()` which takes an enum that indicates if the user wants to use the default http client or one they provide.

Removed eventLoopGroupProvider as the user can provide this via their own httpClient and supplying both an httpClient and an eventLoopGroup doesn't make sense as the only place we use the eventLoopGroup is in the httpClient.

This allows the user to provide a version of `AsyncHTTPClient` with their own configuration, an application global HTTP client, or even a custom written client as long as it conforms to `AWSHTTPClient`.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -34,9 +34,9 @@ public final class AWSClient {
     }
 
     /// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
-    public enum EventLoopGroupProvider {
+    public enum HTTPClientProvider {
         /// `EventLoopGroup` will be provided by the user. Owner of this group is responsible for its lifecycle.
-        case shared(EventLoopGroup)
+        case shared(AWSHTTPClient)
         /// `EventLoopGroup` will be created by the client. When `syncShutdown` is called, created `EventLoopGroup` will be shut down as well.
         case useAWSClientShared
     }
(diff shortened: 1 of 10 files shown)


adam-fowler:
Question: should we provide a default argument for the httpClientProvider. Currently it follows the same setup as eventLoopGroupProvider where the AWSClient doesn't have a default variable, but the service clients do.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -31,11 +31,11 @@ public final class AWSClient {
...
 
     /// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
-    public enum EventLoopGroupProvider {
+    public enum HTTPClientProvider {
         /// `EventLoopGroup` will be provided by the user. Owner of this group is responsible for its lifecycle.
-        case shared(EventLoopGroup)
+        case shared(AWSHTTPClient)
         /// `EventLoopGroup` will be created by the client. When `syncShutdown` is called, created `EventLoopGroup` will be shut down as well.
-        case useAWSClientShared
+        case createNew
```
fabianfett: I think, I prefer if the users must bring their own `HTTPClient`. If we create a new one for them, we will need to shut the client down ourselves. Which leads us to a `shutdown` function. But that is only needed if the users uses `createNew`. All I want to say: it becomes messy.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -62,21 +62,13 @@ public final class AWSClient {
 
     var possibleErrorTypes: [AWSErrorType.Type]
 
-    let httpClient: AWSHTTPClient
+    /// HTTP client used by AWSClient
+    public let httpClient: AWSHTTPClient
 
-    public let eventLoopGroup: EventLoopGroup
+    let httpClientProvider: HTTPClientProvider
```
fabianfett: We wouldn't need this.

Sources/AWSSDKSwiftCore/AWSClient.swift
adam-fowler: There are different HTTP Client requirements for different platforms. Do you expect everyone to know these out of the box? I'm not overly happy with the shutdown process (ie in the deinit) but if we move to the single client I would be happy to move to a syncShutdown() a la AsyncHTTPClient. If the user is required to create their own HTTPClient the shutdown mess is forced onto them.
```
let httpClient = ChooseTheRightClient()
let awsClient = AWSClient(httpClient)
...
httpClient.syncShutdown()
```
vs
```
let awsClient = AWSClient(httpClientProvider: .createNew)
...
awsClient.syncShutdown()
```

PR #206, opened 2020-03-05
Vendor INIParser

adam-fowler:
Also added NOTICE.txt indicating where INIParser came from


adam-fowler:
@fabianfett yep just a copy

fabianfett:
@adam-fowler okay, go ahead and merge!

PR #215, opened 2020-03-17
Properties wrappers for encoding/decoding collections

adam-fowler:
Associated aws-sdk-swift PR is https://github.com/swift-aws/aws-sdk-swift/pull/255

AWS services have many ways to encode collections in XML and querys. This complicates both XMLEncoder/Decoder and QueryEncoder and makes them quite hard to read. This change moves all this code outside of the encoders using property wrappers. There is only one implementation of the collection encoding, instead a copy for each encoder, and it is much easier to read. Also the encoders/decoders look much cleaner.

Previously the collection encoding was included in the AWSMemberEncoding structure. For each `Codable` in a structure the encoder/decoder would search this array to see if it had a collection encoding setup. This changes allows us to do away with this search. It makes the encoders easier to read and more optimal.

See here for more info on Codable property wrappers http://marksands.github.io/2019/10/21/better-codable-through-property-wrappers.html


adam-fowler:
Example shape code with property wrappers. 
```
class MyShape: AWSShape {
    @DictionaryEncoding<DictionaryEntryKeyValue, String, Int> var dictionary: [String: Int]
    @ArrayEncoding<ArrayMember, String> var array[String]
}
```
`DictionaryEntryKeyValue` and `ArrayMember` define the component names. 
```
public struct ArrayMember: ArrayEncodingProperties {
    public static let member = "member"
}
public struct DictionaryEntryKeyValue: DictionaryEncodingProperties {
    public static let entry: String? = "entry"
    public static let key = "key"
    public static let value = "value"
}
```
As these are the most common naming schemes I have added these classes into aws-sdk-swift-core.

adam-fowler:
I can see how I could use propertyWrappers to deal with header and query parameters as well.

adam-fowler:
Due to an issue with optional variables I had to change this quite a bit. There are now two property wrappers. One for non-optional variables (@Coding) and one optional variables (@OptionalCoding). These both have one generic parameter: a coder object conforming to the protocol `CustomCoder`which has two functions one to encode and one to decode. The markup for arrays and dictionaries is now
```
class MyShape: AWSShape {
    @Coding(DictionaryCoder<DictionaryEntryKeyValue, String, Int>> var dictionary: [String: Int]
    @OptionalCoding(ArrayCoder<ArrayMember, String>> var array[String]?
}
```

adam-fowler:
> I can see how I could use propertyWrappers to deal with header and query parameters as well thus allowing us to remove the AWSMemberEncoding array. But I'll leave that for another PR

Unfortunately this may not be possible.

PR #216, opened 2020-03-20
Fix QueryEncoder for Data blob AWSShapeMember

vkill:
Hi, there

The email subject is empty when I send raw email via [SES.sendRawEmail](https://github.com/swift-aws/aws-sdk-swift/blob/4.x.x/Sources/AWSSDKSwift/Services/SES/SES_API.swift#L294), because the request body "RawMessage.Data" is like "xx bytes".

This PR make a AWSShape same as [SES.RawMessage](https://github.com/swift-aws/aws-sdk-swift/blob/4.x.x/Sources/AWSSDKSwift/Services/SES/SES_Shapes.swift#L2199), and tested it.

Please review it. thank you.

### DIFF
--- a/Sources/AWSSDKSwiftCore/Encoder/AWSShapeEncoder.swift
+++ b/Sources/AWSSDKSwiftCore/Encoder/AWSShapeEncoder.swift
@@ -121,7 +121,7 @@ class AWSShapeEncoder {
                         dictionary["\(fullLabel).\(entryString)\(iterator.offset+1).\(valueString)"] = iterator.element.value
                     }
                 case let v as Data:
-                    dictionary[fullLabel] = String(decoding: v, as: UTF8.self)
+                    dictionary[fullLabel] = v.base64EncodedString()
                 default:
                     dictionary[fullLabel] = value
                 }
--- a/Sources/AWSSDKSwiftCore/Encoder/AWSShapeEncoder.swift
+++ b/Sources/AWSSDKSwiftCore/Encoder/AWSShapeEncoder.swift
@@ -120,7 +120,8 @@ class AWSShapeEncoder {
                         dictionary["\(fullLabel).\(entryString)\(iterator.offset+1).\(keyString)"] = iterator.element.key
                         dictionary["\(fullLabel).\(entryString)\(iterator.offset+1).\(valueString)"] = iterator.element.value
                     }
-                
+                case let v as Data:
+                    dictionary[fullLabel] = String(decoding: v, as: UTF8.self)
                 default:
                     dictionary[fullLabel] = value
                 }
--- a/Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
+++ b/Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
@@ -13,7 +13,7 @@ class QueryEncoderTests: XCTestCase {
 
     func queryString(dictionary:[String:Any]) -> String? {
         var components = URLComponents()
-        components.queryItems = dictionary.map( {URLQueryItem(name:$0.key, value:String(describing: $0.value))} ).sorted(by: { $0.name < $1.name })
+        components.queryItems = dictionary.map( {URLQueryItem(name:$0.key, value:String(describing: $0.value).addingPercentEncoding(withAllowedCharacters: AWSClient.queryAllowedCharacters)!)} ).sorted(by: { $0.name < $1.name })
         if components.queryItems != nil, let url = components.url {
             return url.query
         }
@@ -134,7 +134,7 @@ class QueryEncoderTests: XCTestCase {
             let a : Data
         }
         let test = Test(a:Data("Testing".utf8))
-        testQuery(test, query:"A=Testing")
+        testQuery(test, query:"A=VGVzdGluZw%253D%253D")
     }
 
     static var allTests : [(String, (QueryEncoderTests) -> () throws -> Void)] {
--- a/Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
+++ b/Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
@@ -128,6 +128,15 @@ class QueryEncoderTests: XCTestCase {
         testQuery(test, query:"A.1.entry=1&A.1.name=first")
     }
     
+    func testDataBlobEncode() {
+        struct Test : AWSShape {
+            static let _members = [AWSShapeMember(label: "A", required: true, type: .blob)]
+            let a : Data
+        }
+        let test = Test(a:Data("Testing".utf8))
+        testQuery(test, query:"A=Testing")
+    }
+
     static var allTests : [(String, (QueryEncoderTests) -> () throws -> Void)] {
         return [
             ("testSimpleStructureEncode", testSimpleStructureEncode),
@@ -140,6 +149,7 @@ class QueryEncoderTests: XCTestCase {
             ("testArrayEncodingEncode", testArrayEncodingEncode),
             ("testDictionaryEncodingEncode", testDictionaryEncodingEncode),
             ("testDictionaryEncodingEncode2", testDictionaryEncodingEncode2),
+            ("testDataBlobEncode", testDataBlobEncode),
         ]
     }
 }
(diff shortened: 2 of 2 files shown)


adam-fowler:
Hi @vkill ,

thanks for this PR. I'm not sure it is 100% correct though. The comment above RawMessage.data says the data need to be base64 encoded. Have you tested this to send emails?

adam-fowler:
I checked the docs https://docs.aws.amazon.com/ses/latest/APIReference/API_RawMessage.html and yes it should be base64 encoded. So that line should be 
```
                case let v as Data:
                    dictionary[fullLabel] = v.base64EncodedString()
```

Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
```diff
@@ -128,6 +128,15 @@ class QueryEncoderTests: XCTestCase {
         testQuery(test, query:"A.1.entry=1&A.1.name=first")
     }
     
+    func testDataBlobEncode() {
+        struct Test : AWSShape {
+            static let _members = [AWSShapeMember(label: "A", required: true, type: .blob)]
+            let a : Data
+        }
+        let test = Test(a:Data("Testing".utf8))
+        testQuery(test, query:"A=Testing")
```
adam-fowler: If we replace with base64 encoding you will have to test against a URL percent encoded string. So your test would be
```
testQuery(test, query:"A=\(data.base64EncodedString().addingPercentEncoding(withAllowedCharacters: AWSClient.queryAllowedCharacters)!)")
```

vkill:
> Hi @vkill ,
> 
> thanks for this PR. I'm not sure it is 100% correct though. The comment above RawMessage.data says the data need to be base64 encoded. Have you tested this to send emails?

The example

```
let rawMessageString = """
From: from@example.com
Subject: Subject
To: to@example.com
        
        Body
"""
let sesSendRawEmailRequest = SES.SendRawEmailRequest(
    destinations: ["to@example.com"],
    rawMessage: .init(data: Data(rawMessageString.utf8).base64EncodedData()),
    source: "from@example.com"
)

ses.sendRawEmail(sesSendRawEmailRequest)
```

But [awsRequest.body](https://github.com/swift-aws/aws-sdk-swift-core/blob/4.x.x/Sources/AWSSDKSwiftCore/AWSClient.swift#L264) is `"Action=SendRawEmail&Destinations.member.1=to%40example.com&RawMessage.Data=108%20bytes&Source=from%40example.com&Version=2010-12-01"`

vkill:
> I checked the docs https://docs.aws.amazon.com/ses/latest/APIReference/API_RawMessage.html and yes it should be base64 encoded. So that line should be
> 
> ```
>                 case let v as Data:
>                     dictionary[fullLabel] = v.base64EncodedString()
> ```

As core library, I think we should not be care the service logic.

Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
vkill: In QueryEncoderTests, we should not care params encode. In SES, that logic should guarantee by [urlEncodeQueryParams](https://github.com/swift-aws/aws-sdk-swift-core/blob/4.x.x/Sources/AWSSDKSwiftCore/AWSClient.swift#L450)

adam-fowler:
> As core library, I think we should not be care the service logic.

I’m not sure what you are getting at here. But I doubt as a general purpose serialisation method converting Data to a utf8 string is what is expected. I would have thought the standard would be base64. 

And the docs say it should be base64 in the case of SES.sendRawEmail(). Have you tested it with base64?

vkill:
> Have you tested it with base64?

Yes, I was tested it, https://github.com/swift-aws/aws-sdk-swift-core/pull/216#issuecomment-601674302

Maybe replace [public let data: Data](https://github.com/swift-aws/aws-sdk-swift/blob/4.x.x/Sources/AWSSDKSwift/Services/SES/SES_Shapes.swift#L2205) to `public let data: String` is better.

adam-fowler:
I think the `aws-sdk-swift-core` code should be doing the base64 encode. This should be standard for all `Data` objects. The end user shouldn't be doing this. 

I can pretty much guarantee any other service that supplies `Data` objects in a query string does the same.

Sources/AWSSDKSwiftCore/Encoder/AWSShapeEncoder.swift
```diff
@@ -120,7 +120,8 @@ class AWSShapeEncoder {
                         dictionary["\(fullLabel).\(entryString)\(iterator.offset+1).\(keyString)"] = iterator.element.key
                         dictionary["\(fullLabel).\(entryString)\(iterator.offset+1).\(valueString)"] = iterator.element.value
                     }
-                
+                case let v as Data:
+                    dictionary[fullLabel] = String(decoding: v, as: UTF8.self)
```
vkill: > I can pretty much guarantee any other service that supplies Data objects in a query string requires them to be base64 encoded as well.

So, just replace this line to `dictionary[fullLabel] = v.base64EncodedString()` and test it?

adam-fowler: Yep. I believe that is the correct way to do this.

vkill: I see, I will change it later.

Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
adam-fowler: You are right we shouldn't be doing the percent coding. Perhaps the test should be as follows given the function `testQuery` uses URLComponents to bulid the query string
```
        var components = URLComponents()
        components.queryItems = [URLQueryItem(name: "A", value: data.base64EncodedString())]
        testQuery(test, query: components.url!.query!)
```

Sources/AWSSDKSwiftCore/Encoder/AWSShapeEncoder.swift
vkill: @adam-fowler I have modified, please review again.

Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
```diff
@@ -13,7 +13,7 @@ class QueryEncoderTests: XCTestCase {
 
     func queryString(dictionary:[String:Any]) -> String? {
         var components = URLComponents()
-        components.queryItems = dictionary.map( {URLQueryItem(name:$0.key, value:String(describing: $0.value))} ).sorted(by: { $0.name < $1.name })
+        components.queryItems = dictionary.map( {URLQueryItem(name:$0.key, value:String(describing: $0.value).addingPercentEncoding(withAllowedCharacters: AWSClient.queryAllowedCharacters)!)} ).sorted(by: { $0.name < $1.name })
```
adam-fowler: Why have you added this line. You are percent encoding the text twice. `URLComponents` percent encodes the url already.

Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
```diff
@@ -128,6 +128,15 @@ class QueryEncoderTests: XCTestCase {
         testQuery(test, query:"A.1.entry=1&A.1.name=first")
     }
     
+    func testDataBlobEncode() {
+        struct Test : AWSShape {
+            static let _members = [AWSShapeMember(label: "A", required: true, type: .blob)]
+            let a : Data
+        }
+        let test = Test(a:Data("Testing".utf8))
+        testQuery(test, query:"A=VGVzdGluZw%253D%253D")
```
adam-fowler: We want to test that the QueryEncoder returns the correct dictionary. Create the dictionary you would expect and then create a string from it as we for the queries and use that as your comparison
```swift
let result = ["A": Data("Testing".utf8).base64EncodedString()]
testQuery(test, query: queryString(dictionary: result))
```

Tests/AWSSDKSwiftCoreTests/QueryEncoderTests.swift
vkill: Oh, it's my mistake, I have removed it.

adam-fowler:
Ok this looks good.

adam-fowler:
@vkill Thank you for this

PR #218, opened 2020-03-20
Payload Object

adam-fowler:
Associated PR in aws-sdk-swift is https://github.com/swift-aws/aws-sdk-swift/pull/257

Add a `AWSPayload` object for data blob payloads in requests. The internal representation is ByteBuffer but allow it to be initialised with `Data`, `ByteBuffer` or `String`. While the user still has the ability to create a payload from a `Data` object as before, they can also supply a `ByteBuffer` which they may have received from another process eg `NIONonBlockingFileIO`.

NB This also removes `Body.init(anyValue: Any)`


adam-fowler:
`AWSPayload` currently conforms to `Codable`. While it doesn't require the encode/decode functions to be implemented it does need conformance to be included as a member of an AWSShape. Once #214 is merged this requirement is removed as aren't placing this in response shapes and we can remove the coding key for the payload member variable in any requests.

Issue #226, opened 2020-04-02
Allow AWSClient Region to be configured with an arbitrary string


clenart:
**Is your feature request related to a problem? Please describe.**
When interacting with non-AWS services which provide an implementation of an AWS API, it can be necessary to configure a Region string for V4 signing, which is not in the set of Regions that AWS provides.

**Describe the solution you'd like**
While the Region enum is user friendly and what 99% of clients will likely need, it would be useful to represent the Region internally as a String. And provide an init or setter method, which allows the Region to be set as any arbitrary string.

**Describe alternatives you've considered**
I'm not sure there is any other way to handle it. If arbitrary strings are not allowed in the V4 Region name, then this library cannot interact with other cloud providers who implement AWS APIs (S3 API being the most popular, for example).

**Additional context**

adam-fowler:
Which cloud providers are you thinking?

PR #228, opened 2020-04-03
Added possibility of arbituary named regions

adam-fowler:
Allow users to provide an arbituary region name using enum case Region.other(String). See #226

This has meant we have had to add our own version of `init(rawValue)` and `rawValue`.


adam-fowler:
@Yasumoto this could resolve https://github.com/swift-aws/aws-sdk-swift/issues/143. What's your thoughts

PR #237, opened 2020-04-15
Add support for streaming request payload

adam-fowler:
AWSPayload is now an enum and has a `.stream` case which has parameters `size` and `closure`. `closure` supplies `EventLoopFuture<ByteBuffer>`. It is repeatedly called until enough data has been supplied.

Internally `Body.buffer` is now `Body.raw` and has a AWSPayload as a parameter. As with the streaming response PR this functionality is only available with AsyncHTTPClient.

TODO:
- [x] Check what to do for signing of requests that aren't S3

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
+++ b/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
@@ -45,20 +45,33 @@ public enum AWSPayload {
 
     /// return payload as Data
     public func asData() -> Data? {
-        return byteBuffer.getData(at: byteBuffer.readerIndex, length: byteBuffer.readableBytes, byteTransferStrategy: .noCopy)
+        switch self {
+        case .byteBuffer(let byteBuffer):
+            return byteBuffer.getData(at: byteBuffer.readerIndex, length: byteBuffer.readableBytes, byteTransferStrategy: .noCopy)
+        default:
+            return nil
+        }
     }
 
     /// return payload as String
     public func asString() -> String? {
-        return byteBuffer.getString(at: byteBuffer.readerIndex, length: byteBuffer.readableBytes, encoding: .utf8)
+        switch self {
+        case .byteBuffer(let byteBuffer):
+            return byteBuffer.getString(at: byteBuffer.readerIndex, length: byteBuffer.readableBytes, encoding: .utf8)
+        default:
+            return nil
+        }
     }
 
     /// return payload as ByteBuffer
-    public func asBytebuffer() -> ByteBuffer {
-        return byteBuffer
+    public func asByteBuffer() -> ByteBuffer? {
+        switch self {
+        case .byteBuffer(let byteBuffer):
+            return byteBuffer
+        default:
+            return nil
+        }
     }
-
-    let byteBuffer: ByteBuffer
 }
 
 extension AWSPayload: Decodable {
--- a/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
+++ b/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
@@ -33,7 +33,8 @@ public struct AWSPayload {
         return AWSPayload(payload: .byteBuffer(buffer))
     }
     
-    /// construct a payload from a stream function
+    /// construct a payload from a stream function. If you supply a size the stream function will be called repeated until you supply the number of bytes specified. If you
+    /// don't supply a size the stream function will be called repeatedly until you supply an empty `ByteBuffer`
     public static func stream(size: Int? = nil, stream: @escaping (EventLoop)->EventLoopFuture<ByteBuffer>) -> Self {
         return AWSPayload(payload: .stream(size: size, stream: stream))
     }
--- a/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
+++ b/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
@@ -21,6 +21,7 @@ public enum AWSPayload {
     case byteBuffer(ByteBuffer)
     case stream(size: Int, closure: (EventLoop)->EventLoopFuture<ByteBuffer>)
 
+    /// Construct a payload from a Data
     public static func data(_ data: Data) -> Self {
         var byteBuffer = ByteBufferAllocator().buffer(capacity: data.count)
         byteBuffer.writeBytes(data)
@@ -32,6 +33,24 @@ public enum AWSPayload {
         var byteBuffer = ByteBufferAllocator().buffer(capacity: string.utf8.count)
         byteBuffer.writeString(string)
         return .byteBuffer(byteBuffer)
+    }
+
+    /// Construct a payload from a NIOFileHandle
+    public static func fileHandle(_ fileHandle: NIOFileHandle, size: Int, fileIO: NonBlockingFileIO, byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()) -> Self {
+        let blockSize = 64*1024
+        var leftToRead = size
+        func stream(_ eventLoop: EventLoop) -> EventLoopFuture<ByteBuffer> {
+            let blockSize = min(leftToRead, blockSize)
+            leftToRead -= blockSize
+            let futureByteBuffer: EventLoopFuture<ByteBuffer> = fileIO.read(fileHandle: fileHandle, byteCount: blockSize, allocator: byteBufferAllocator, eventLoop: eventLoop)
+                .map { byteBuffer in
+                    precondition(byteBuffer.readableBytes == blockSize, "File did not have enough data")
+                    return byteBuffer
+            }
+            return futureByteBuffer
+        }
+        
+        return .stream(size: size, closure: stream)
     }
 
     var size: Int {
--- a/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
+++ b/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
@@ -35,7 +35,7 @@ public enum AWSPayload {
         return .byteBuffer(byteBuffer)
     }
 
-    /// Construct a payload from a NIOFileHandle
+    /// Construct a stream payload from a NIOFileHandle
     public static func fileHandle(_ fileHandle: NIOFileHandle, size: Int, fileIO: NonBlockingFileIO, byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()) -> Self {
         let blockSize = 64*1024
         var leftToRead = size
@@ -76,7 +76,7 @@ public enum AWSPayload {
     public func asString() -> String? {
         switch self {
         case .byteBuffer(let byteBuffer):
-            return byteBuffer.getString(at: byteBuffer.readerIndex, length: byteBuffer.readableBytes, encoding: .utf8)
+            return byteBuffer.getString(at: byteBuffer.readerIndex, length: byteBuffer.readableBytes)
         default:
             return nil
         }
--- a/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
+++ b/Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
@@ -16,26 +16,43 @@ import struct Foundation.Data
 import NIO
 import NIOFoundationCompat
 
-public enum AWSPayload {
-
-    case byteBuffer(ByteBuffer)
-    case stream(size: Int? = nil, stream: (EventLoop)->EventLoopFuture<ByteBuffer>)
-
-    /// Construct a payload from a Data
+/// Holds a request or response payload. A request payload can be in the form of either a ByteBuffer or a stream function that will supply ByteBuffers to the HTTP client.
+/// A response payload only comes in the form of a ByteBuffer
+public struct AWSPayload {
+    
+    /// Internal enum
+    enum Payload {
+        case byteBuffer(ByteBuffer)
+        case stream(size: Int?, stream: (EventLoop)->EventLoopFuture<ByteBuffer>)
+    }
+    
+    internal let payload: Payload
+    
+    /// construct a payload from a ByteBuffer
+    public static func byteBuffer(_ buffer: ByteBuffer) -> Self {
+        return AWSPayload(payload: .byteBuffer(buffer))
+    }
+    
+    /// construct a payload from a stream function
+    public static func stream(size: Int? = nil, stream: @escaping (EventLoop)->EventLoopFuture<ByteBuffer>) -> Self {
+        return AWSPayload(payload: .stream(size: size, stream: stream))
+    }
+    
+    /// Construct a payload from `Data`
     public static func data(_ data: Data) -> Self {
         var byteBuffer = ByteBufferAllocator().buffer(capacity: data.count)
         byteBuffer.writeBytes(data)
-        return .byteBuffer(byteBuffer)
+        return AWSPayload(payload: .byteBuffer(byteBuffer))
     }
 
-    /// Construct a payload from a String
+    /// Construct a payload from a `String`
     public static func string(_ string: String) -> Self {
         var byteBuffer = ByteBufferAllocator().buffer(capacity: string.utf8.count)
         byteBuffer.writeString(string)
-        return .byteBuffer(byteBuffer)
+        return AWSPayload(payload: .byteBuffer(byteBuffer))
     }
 
-    /// Construct a stream payload from a NIOFileHandle
+    /// Construct a stream payload from a `NIOFileHandle`
     public static func fileHandle(_ fileHandle: NIOFileHandle, size: Int? = nil, fileIO: NonBlockingFileIO, byteBufferAllocator: ByteBufferAllocator = ByteBufferAllocator()) -> Self {
         let blockSize = 64*1024
         var leftToRead = size
@@ -57,11 +74,12 @@ public enum AWSPayload {
             return futureByteBuffer
         }
         
-        return .stream(size: size, stream: stream)
+        return AWSPayload(payload: .stream(size: size, stream: stream))
     }
-
+    
+    /// Return the size of the payload. If the payload is a stream it is always possible to return a size
     var size: Int? {
-        switch self {
+        switch payload {
         case .byteBuffer(let byteBuffer):
             return byteBuffer.readableBytes
         case .stream(let size, _):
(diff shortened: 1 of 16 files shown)


adam-fowler:
`MediaStoreData.PutObject` and `LexRuntimeService.PostContent` both can be uploaded with an unsigned body. The operations are flagged with `"authtype":"v4-unsigned-body"` in the json model files

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
```diff
@@ -18,13 +18,53 @@ import NIO
...
-            requestBody = .byteBuffer(body)
+            switch body {
+            case .byteBuffer(let byteBuffer):
+                requestBody = .byteBuffer(byteBuffer)
+
+            case .stream(let size, let closure):
+                let eventLoop = eventLoopGroup.next()
+                requestBody = .stream(length: size) { writer in
+                    return self.writeToStreamWriter(writer: writer, size: size, on: eventLoop, closure: closure)
+                }
```
fabianfett, line 66: With the proposed `FileRegion` addition:

```swift
case .fileRegion(let region):
    writer.write(.fileRegion(region))
```

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
```diff
@@ -18,13 +18,53 @@ import NIO
...
         if let eventLoop = eventLoop {
             precondition(self.eventLoopGroup.makeIterator().contains { $0 === eventLoop }, "EventLoop provided to AWSClient must be part of the HTTPClient's EventLoopGroup.")
-        }
+        }        
         let requestBody: AsyncHTTPClient.HTTPClient.Body?
         if let body = request.body {
-            requestBody = .byteBuffer(body)
+            switch body {
+            case .byteBuffer(let byteBuffer):
+                requestBody = .byteBuffer(byteBuffer)
```
fabianfett, line 60: Do we maybe want to combine this? Will save us one level of indention:
```swift
switch request.body {
case .some(.byteBuffer):
    requestBody = .byteBuffer(byteBuffer)
case .some(.stream(let closure)):
    ...
case .some(.stream(let closure)):
    ...
case .none:
    requestBody = nil
}
```

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
```diff
@@ -100,23 +100,39 @@ public struct AWSRequest {
...
-        return AWSHTTPRequest.init(url: signedURL, method: method, headers: getHttpHeaders(), body: bodyData)
+        let payload = self.body.asPayload()
+        let bodyDataForSigning: AWSSigner.BodyData?
+        if let body = payload?.asByteBuffer() {
+            bodyDataForSigning = .byteBuffer(body)
+        } else if case .stream = payload {
+            bodyDataForSigning = .unsignedPayload
+        } else {
+            bodyDataForSigning = nil
+        }
```
fabianfett, line 117: use a switch?

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
```diff
@@ -16,44 +16,81 @@ import struct Foundation.Data
...
-/// Object storing request/response payload
-public struct AWSPayload {
+public enum AWSPayload {
 
-    /// Construct a payload from a ByteBuffer
-    public static func byteBuffer(_ byteBuffer: ByteBuffer) -> Self {
-        return Self(byteBuffer: byteBuffer)
-    }
+    case byteBuffer(ByteBuffer)
+    case stream(size: Int, closure: (EventLoop)->EventLoopFuture<ByteBuffer>)
```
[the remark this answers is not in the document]
adam-fowler: Fair point about the enum, will convert to a struct
Unfortunately the stream size is required by AHC. There is an issue about this https://github.com/swift-server/async-http-client/issues/194
I also added a request for adding FileRegion to AHC but then removed it as FileRegion only works well for http. For https they would have to load the file to encrypt it. See conversation https://github.com/swift-server/async-http-client/pull/202

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
```diff
@@ -16,44 +16,81 @@ import struct Foundation.Data
...
+        let blockSize = 64*1024
+        var leftToRead = size
+        func stream(_ eventLoop: EventLoop) -> EventLoopFuture<ByteBuffer> {
+            let blockSize = min(leftToRead, blockSize)
+            leftToRead -= blockSize
+            let futureByteBuffer: EventLoopFuture<ByteBuffer> = fileIO.read(fileHandle: fileHandle, byteCount: blockSize, allocator: byteBufferAllocator, eventLoop: eventLoop)
+                .map { byteBuffer in
+                    precondition(byteBuffer.readableBytes == blockSize, "File did not have enough data")
+                    return byteBuffer
+            }
```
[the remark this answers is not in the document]
adam-fowler: If you are talking about the `precondition`. It is something we have to check. It is checking the user has passed us a size that the file is able to deliver. If we ask for 64k, because user tells us there is 64k, and the file is only 32k then we get a bytebuffer of a 32k. AHC is left waiting for the remaing 32k but we keep passing empty bytebuffers to AHC.

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
adam-fowler: ok

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
adam-fowler: See comment above

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
adam-fowler: yeah not sure why that isnt a switch will sort

Sources/AWSSDKSwiftCore/HTTP/NIOTSHTTPClient.swift
```diff
@@ -254,7 +254,18 @@ extension NIOTSHTTPClient: AWSHTTPClient {
...
+        if let body = request.body {
+            switch body {
+            case .byteBuffer(let byteBuffer):
+                requestBody = byteBuffer
+            case .stream:
+                preconditionFailure("Request streaming isnt supported")
+            }
+        } else {
+            requestBody = nil
+        }
```
fabianfett, line 267:
```suggestion
        switch request.body {   
        case .some(.byteBuffer(let byteBuffer)):
            requestBody = byteBuffer
        case .some(.stream):
            preconditionFailure("Request streaming isnt supported")
        default:
            requestBody = nil
        }
```

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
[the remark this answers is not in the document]
adam-fowler: AWSPayload is now a struct and the stream size is optional.

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
adam-fowler: done

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
adam-fowler: done

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
```diff
@@ -16,44 +16,106 @@ import struct Foundation.Data
...
 public struct AWSPayload {
-
-    /// Construct a payload from a ByteBuffer
-    public static func byteBuffer(_ byteBuffer: ByteBuffer) -> Self {
-        return Self(byteBuffer: byteBuffer)
+    
+    /// Internal enum
+    enum Payload {
+        case byteBuffer(ByteBuffer)
+        case stream(size: Int?, stream: (EventLoop)->EventLoopFuture<ByteBuffer>)
```
fabianfett, line 26: move size to the back and make an `UInt?`

fabianfett: what do you think about adding a `.none` case? that way we don't need to make `AWSPayload` an optional everywhere.

adam-fowler: Given most people are going to define a stream as follows
```
let payload = .stream(size:16384) {
   ...
}
```
I'm not sure the order matters
Also no point setting the size to `UInt` when the size in AHC is an `Int`

Sources/AWSSDKSwiftCore/HTTP/NIOTSHTTPClient.swift
adam-fowler: done

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
adam-fowler: `Int.max` would support a stream size of 9 exabytes.

adam-fowler: Regarding the `.none` payload (I prefer `.empty`). Yeah I can look into that. I'll add an issue to look into after though.

fabianfett: I'm not sure if it is a good idea to postpone `empty` until a next pr. we change the AWSPayload in this pr anyway. I'm not sure, if we want two half changes?

Your call... Just raising the issue!

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
```diff
@@ -16,44 +16,107 @@ import struct Foundation.Data
...
+    }
+    
+    /// construct a payload from a stream function. If you supply a size the stream function will be called repeated until you supply the number of bytes specified. If you
+    /// don't supply a size the stream function will be called repeatedly until you supply an empty `ByteBuffer`
+    public static func stream(size: Int? = nil, stream: @escaping (EventLoop)->EventLoopFuture<ByteBuffer>) -> Self {
+        return AWSPayload(payload: .stream(size: size, stream: stream))
+    }
+    
+    /// Construct a payload from `Data`
     public static func data(_ data: Data) -> Self {
```
fabianfett, line 43: I think we should give the option to inject a `ByteBufferAllocator` here. Default value is create a new one.

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
```diff
@@ -16,44 +16,107 @@ import struct Foundation.Data
...
         var byteBuffer = ByteBufferAllocator().buffer(capacity: data.count)
         byteBuffer.writeBytes(data)
-        return Self(byteBuffer: byteBuffer)
+        return AWSPayload(payload: .byteBuffer(byteBuffer))
     }
 
-    /// Construct a payload from a String
+    /// Construct a payload from a `String`
     public static func string(_ string: String) -> Self {
```
fabianfett, line 50: I think we should give the option to inject a `ByteBufferAllocator` here. Default value is create a new one.

Sources/AWSSDKSwiftCore/Doc/AWSShape.swift
```diff
@@ -106,10 +106,14 @@ public extension AWSEncodableShape {
         guard value.count <= max else { throw AWSClientError(.validationError, message: "Length of \(parent).\(name) (\(value.count)) is greater than the maximum allowed value \(max).") }
     }
     func validate(_ value: AWSPayload, name: String, parent: String, min: Int) throws {
-        guard value.byteBuffer.readableBytes >= min else { throw AWSClientError(.validationError, message: "Length of \(parent).\(name) (\(value.byteBuffer.readableBytes)) is less than minimum allowed value \(min).") }
+        if let size = value.size {
+            guard size >= min else { throw AWSClientError(.validationError, message: "Length of \(parent).\(name) (\(size)) is less than minimum allowed value \(min).") }
```
fabianfett, line 110: could we break after else for better readability?

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
```diff
@@ -18,19 +18,71 @@ import NIO
 
 /// comply with AWSHTTPClient protocol
 extension AsyncHTTPClient.HTTPClient: AWSHTTPClient {
+
+    /// write stream to StreamWriter
+    private func writeToStreamWriter(
+        writer: HTTPClient.Body.StreamWriter,
+        size: Int?,
+        on eventLoop: EventLoop,
+        stream: @escaping (EventLoop)->EventLoopFuture<ByteBuffer>) -> EventLoopFuture<Void> {
```
fabianfett, line 27: this is not really a stream. it is more a get more data callback. do we maybe want to name it. `getMoreData`.

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
```diff
@@ -18,19 +18,71 @@ import NIO
...
+            _ = stream(eventLoop)
+                .map { (byteBuffer)->() in
+                    // if no amount was set and the byte buffer has no readable bytes then this is assumed to mean
+                    // there will be no more data
+                    if amountLeft == nil && byteBuffer.readableBytes == 0 {
+                        promise.succeed(())
+                        return
+                    }
+                    let newAmountLeft = amountLeft.map { $0 - byteBuffer.readableBytes }
+                    _ = writer.write(.byteBuffer(byteBuffer)).flatMap { ()->EventLoopFuture<Void> in
```
fabianfett, line 42: we will need to talk to Cory here, if it reasonable to wait after every write. Or if we should write a number of bytes as fast as possible and then wait till those are written to disk. Maybe we write 16kb as fast as we can and then we wait for the buffer to be empty. otherwise with small byte sizes this might get very slow.

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
```diff
@@ -18,19 +18,71 @@ import NIO
...
-        } else {
+        var requestHeaders = request.headers
+        
+        switch request.body?.payload {
+        case .some(.byteBuffer(let byteBuffer)):
+            requestBody = .byteBuffer(byteBuffer)
+        case .some(.stream(let size, let closure)):
+            // add "Transfer-Encoding" header if streaming with unknown size
+            if size == nil {
+                requestHeaders.add(name: "Transfer-Encoding", value: "chunked")
```
fabianfett, line 76: awesome. if we do this though we need to add some kind of magic headers between the chunks. i think those are missing right now.

adam-fowler: With an unsigned-payload stream you don't need the magic headers. There are a number of services that only accept unsigned-payload streams, also S3 does accept unsigned-payloads. I have already added an issue to do a signed stream #248 at a later date.

Also if you don't provide a size it isn't possible to do a signed streamed upload as the final content-size is required.

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
adam-fowler: You need to wait, to avoid back pressure. There is related conversation here https://github.com/swift-server/async-http-client/issues/194#issuecomment-629249604. We could collate bytebuffers until we reach a certain size, but that does complicate things considerably. Maybe for another PR?

fabianfett: Okay. Later pr sounds good to me.

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
fabianfett: Does ahc add the sizes and \r\n between the chunks as defined here:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding

adam-fowler: @fabianfett good point, will need to check. So far I've only checked this with non-chunked uploads.

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
adam-fowler: I can do, just trying to keep PR smallish. I guess this isnt a huge addition

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
adam-fowler: yep it failed doing an S3.PutObject. This is something I would have thought was really up to AHC, but then they expect you to set the "chunked" header. I'll ask them

adam-fowler: https://github.com/swift-server/async-http-client/issues/212

adam-fowler: By the way AWS have their own version of this called "aws-chunked" which includes signing information in it. https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
adam-fowler: `getData` should be enough, `getMoreData` implies we've already had some

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
adam-fowler: done

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
adam-fowler: done

Sources/AWSSDKSwiftCore/Doc/AWSShape.swift
adam-fowler: done

Sources/AWSSDKSwiftCore/HTTP/AsyncHTTPClient.swift
adam-fowler: done

Sources/AWSSDKSwiftCore/AWSShapes/Payload.swift
adam-fowler: done

PR #238, opened 2020-04-16
Enable JitterRetry RetryController by default

adam-fowler:
Currently the retry code submitted in #202 is disabled. This PR enables the code by defaulting to using the JitterRetry controller. 

The Jitter retry calculates the wait time between retry requests in the recommended way from AWS. It adds a random element to the retry time calculation to ensure that all the requests that failed at the same time do much keep retrying at the same time.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -62,7 +62,7 @@ public final class AWSClient {
     /// EventLoopGroup used by AWSClient
     public var eventLoopGroup: EventLoopGroup { return httpClient.eventLoopGroup }
     /// Retry Controller specifying what to do when a request fails
-    public let retryStrategy: RetryStrategy
+    public let retryPolicy: RetryPolicy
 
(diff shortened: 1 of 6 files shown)


fabianfett:
I really like this. The only thing i‘m not sure about is the controller name part in retrycontroller. But i cant come up with anything better that isn‘t totally generic.

adam-fowler:
@fabianfett The one thing I'm not keen on is the return value of `getRetryWaitTime` where `nil` means dont retry. What do you think of changing that from an optional `TimeAmount` to an enum
```
enum RetryStatus {
    case retry(wait: TimeAmount)
    case dontRetry
}
```

adam-fowler:
> I really like this. The only thing i‘m not sure about is the controller name part in retrycontroller. But i cant come up with anything better that isn‘t totally generic.

Yeah I struggled with a name, this was the best I could come up with

fabianfett:
In go the protocol is called `RequestRetryer`. I think i like that! @adam-fowler wdyt?
https://github.com/aws/aws-sdk-go/blob/d7b5372b5b7583d5a30a3d7440d9d684aba581a9/aws/config.go#L108

PR #240, opened 2020-04-16
Added `on eventLoop: EventLoop` parameter to send() functions

adam-fowler:
Added `on eventLoop: EventLoop` parameter to send() functions and then pass this eventLoop through all the systems.

This allows users of aws-sdk-swift more control over which eventloop their requests run on.


adam-fowler:
> lg. this will require an update to the main sdk as well yeah?
> 
Here is the aws-sdk-swift PR https://github.com/swift-aws/aws-sdk-swift/pull/273.

PR #246, opened 2020-05-11
Make Region and Partition structs so we can extend them

adam-fowler:
Make Region and Partition structs so we can extend them without having to up the major version number.

Yuck I hate this but it is necessary


Sources/AWSSDKSwiftCore/Doc/Region.swift
```diff
@@ -14,151 +14,142 @@
...
+/*extension Region: Equatable, Hashable {
     public static func == (lhs: Region, rhs: Region) -> Bool {
         lhs.rawValue == rhs.rawValue
     }
 
     public func hash(into hasher: inout Hasher) {
         self.rawValue.hash(into: &hasher)
     }
-}
+}*/
```
fabianfett: remove?

PR #262, opened 2020-05-21
Removed `AWSClient.signURL`

fabianfett:
Removed `AWSClient.signURL` that is just a proxy to the `AWSSigner`. Keep the API clean. With 5.0.0 we can remove it!

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -341,17 +341,6 @@ extension AWSClient {
         }
         return recordMetrics(future, service: serviceConfig.service, operation: operationName)
     }
-
-    /// generate a signed URL
-    /// - parameters:
-    ///     - url : URL to sign
-    ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
-    ///     - expires: How long before the signed URL expires
-    /// - returns:
-    ///     A signed URL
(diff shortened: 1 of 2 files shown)


adam-fowler:
I think this should stay. It is something I get asked about all the time, and if we remove it, it will make it quite hard to find the functionality. Having a helper function for popular features is not necessarily indicative of an untidy api.

fabianfett:
@adam-fowler I would like to argue that we should expose `AWSSignerV4` as a library product. Users can use the signer directly then.

adam-fowler:
> @adam-fowler I would like to argue that we should expose `AWSSignerV4` as a library product. Users can use the signer directly then. 

Funny you were arguing for it to be implementation only a few months back :-)

I’m adding some public symbols to it for the s3 chunked signing which don’t make sense outside of aws-sdk-swift-core so not sure about it being a library product.

fabianfett:
> I’m adding some public symbols to it for the s3 chunked signing which don’t make sense outside of aws-sdk-swift-core so not sure about it being a library product.

Could those be extensions within `AWSSDKSwiftCore`?

adam-fowler:
If we remove the `signURL` function we'll need to add a function to provide a signer to the user, so they can call the signer functions.

The s3 chunk signing stuff uses internal symbols from AWSSignerV4, so moving it to AWSSDKSwiftCore would require exposing more of AWSSignerV4. I would suggest we remove `AWSSignerV4` from exports.swift (shouldnt really be there) and leave the one `signURL` function in AWSClient.

PR #271, opened 2020-05-21
Move TestServer and TestUtils to separate AWSTestUtils target

adam-fowler:
- Cleaned up the AWSTestServer at the same time
- Made XMLEncoder/XMLDecoder public as the AWSTestServer uses these and any tools that used it would require them.


Sources/AWSSDKSwiftCore/Encoder/XMLDecoder.swift
```diff
@@ -18,7 +18,7 @@ import struct Foundation.URL
 import class  Foundation.DateFormatter
 
 /// The wrapper class for decoding Codable classes from XMLNodes
-class XMLDecoder {
+public class XMLDecoder {
```
fabianfett, line 21: we could move this to a separate `AWSXML` target and don't export this as a `library` product. this way we don't expose this to end users and we would keep our api smaller (and less breaking).

adam-fowler:
I'm going to merge this, given the XML move is being done in another PR

fabianfett:
@adam-fowler the other pr would merge into this branch!

adam-fowler:
oh alright fair enough. We'll merge that one first

Issue #266, opened 2020-05-21
Always sign request by adding Header fields


fabianfett:
Right now we try to sign `HEAD` and `GET` request by adding the signing parameters to the URL. According the [AWS documentation](https://docs.aws.amazon.com/general/latest/gr/sigv4-add-signature-to-request.html) there is no difference in signing urls vs headers. 

Since signed urls are generally something that should be handled with extra care, we should always sign by adding signing headers.

The code to be changed can be found in `AWSClient`:

```swift
    func createHTTPRequest(signer: AWSSigner) -> AWSHTTPRequest {
        // if credentials are empty don't sign request
        if signer.credentials.isEmpty() {
            return self.toHTTPRequest()
        }

        switch (self.httpMethod, self.serviceProtocol) {
        case ("GET",  .restjson), ("HEAD", .restjson):
            return self.toHTTPRequestWithSignedHeader(signer: signer)

        case ("GET",  _), ("HEAD", _):
            if self.httpHeaders.count > 0 {
                return self.toHTTPRequestWithSignedHeader(signer: signer)
            }
            return self.toHTTPRequestWithSignedURL(signer: signer)

        default:
            return self.toHTTPRequestWithSignedHeader(signer: signer)
        }
    }
```

fabianfett:
thanks @Ro-M!

PR #272, opened 2020-05-21
Always sign request by adding Header fields

Ro-M:
- Always sign requests by adding header fields instead of signing the url
- Added a test to check that a signed header is used

### DIFF
--- a/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
+++ b/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
@@ -406,6 +406,35 @@ class AWSClientTests: XCTestCase {
             XCTFail(error.localizedDescription)
         }
     }
+    
+    func testSignedClient() {
+        let input = E()
+        let client = AWSClient(
+            accessKeyId: "foo",
+            secretAccessKey: "bar",
+            region: .useast1,
+            service: "s3",
+            serviceProtocol: .restxml,
+            apiVersion: "2013-12-02",
+            middlewares: [],
+            httpClientProvider: .createNew
+        )
+
+        do {
+            let awsRequest = try client.createAWSRequest(
+                operation: "CopyObject",
+                path: "/",
+                httpMethod: "PUT",
+                input: input
+            )
+
+            let request: AWSHTTPRequest = awsRequest.createHTTPRequest(signer: try client.signer.wait())
+
+            XCTAssertNotNil(request.headers["Authorization"].first)
+        } catch {
+            XCTFail(error.localizedDescription)
+        }
+    }
 
     func testProtocolContentType() throws {
         struct Object: AWSEncodableShape {
--- a/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
+++ b/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
@@ -409,17 +409,9 @@ class AWSClientTests: XCTestCase {
     
     func testSignedClient() {
         let input = E()
-        let client = AWSClient(
-            accessKeyId: "foo",
-            secretAccessKey: "bar",
-            region: .useast1,
-            service: "s3",
-            serviceProtocol: .restxml,
-            apiVersion: "2013-12-02",
-            middlewares: [],
-            httpClientProvider: .createNew
-        )
-
+        
+        let client = createAWSClient(accessKeyId: "foo", secretAccessKey: "bar")
+        
         do {
             let awsRequest = try client.createAWSRequest(
                 operation: "CopyObject",
--- a/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
+++ b/Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
@@ -409,20 +409,21 @@ class AWSClientTests: XCTestCase {
     
     func testSignedClient() {
         let input = E()
-        
         let client = createAWSClient(accessKeyId: "foo", secretAccessKey: "bar")
         
         do {
-            let awsRequest = try client.createAWSRequest(
-                operation: "CopyObject",
-                path: "/",
-                httpMethod: "PUT",
-                input: input
-            )
-
-            let request: AWSHTTPRequest = awsRequest.createHTTPRequest(signer: try client.signer.wait())
-
-            XCTAssertNotNil(request.headers["Authorization"].first)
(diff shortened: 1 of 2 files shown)


Ro-M:
relates to issue #266

Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
```diff
@@ -406,6 +406,35 @@ class AWSClientTests: XCTestCase {
             XCTFail(error.localizedDescription)
         }
     }
+    
+    func testSignedClient() {
+        let input = E()
+        let client = AWSClient(
```
adam-fowler, line 412: I recently added a helper function `createAWSClient()` which fills out all the parameters with default values. If you could use that instead that'd be great. Unless you want a specfic parameter set you can call `createAWSClient()` with nothing set. Means we have less client creation code filling up the tests.

Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
```diff
@@ -406,6 +406,35 @@ class AWSClientTests: XCTestCase {
...
+            region: .useast1,
+            service: "s3",
+            serviceProtocol: .restxml,
+            apiVersion: "2013-12-02",
+            middlewares: [],
+            httpClientProvider: .createNew
+        )
+
+        do {
+            let awsRequest = try client.createAWSRequest(
```
adam-fowler, line 424: Can you try this with different requests, maybe a GET request as well, seeing as that was originally the requests that used SignedURL.

Ro-M: good point. would you like separate tests for that or is it ok if I test multiple requests in a single test?

Ro-M:
on a side note: the `toHTTPRequestWithSignedURL` method on `AWSRequet` is now unused code and not public. Should I remove this method or should it remain for now?

Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
adam-fowler: It's ok to put both of those in the same test

jonnymacs:
@adam-fowler @Ro-M hey guys, this is triggering some alerts for me. I believe signed headers don't work to auth head requests to s3 when running in a container that uses the IAM metadata credential service with a token. I think i lost a bit of my mind making this discovery.

please test that use case.

adam-fowler:
> @adam-fowler @Ro-M hey guys, this is triggering some alerts for me. I believe signed headers don't work to auth head requests to s3 when running in a container that uses the IAM metadata credential service with a token. I think i lost a bit of my mind making this discovery.
> 
> please test that use case.

Seems a very odd failure. AWS do say everything works with headers, but we'll check before merging. @Ro-M you thought you had chosen an easy task as well.

adam-fowler:
> on a side note: the `toHTTPRequestWithSignedURL` method on `AWSRequet` is now unused code and not public. Should I remove this method or should it remain for now?

Yes, if everything works

adam-fowler:
@jonnymacs is that an ECS container? Or a docker image running on EC2?

jonnymacs:
Ecs container.

jonnymacs:
Yeah it was an odd and miserable failure to debug. 401 unauthorized, but
only when the token is required. It was about a year ago, there is a commit
that ensures HEAD request use the signedURL instead of headers that I made
to get it working.

Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
```diff
@@ -406,6 +406,35 @@ class AWSClientTests: XCTestCase {
...
+                operation: "CopyObject",
+                path: "/",
+                httpMethod: "PUT",
+                input: input
+            )
+
+            let request: AWSHTTPRequest = awsRequest.createHTTPRequest(signer: try client.signer.wait())
+
+            XCTAssertNotNil(request.headers["Authorization"].first)
```
fabianfett: if you just want to ensure no error is thrown, we (following @weissi's wisdom 😉) normally use a pattern like this: 

```swift
var request: AWSHTTPRequest?
XCTAssertNoThrow(request = awsRequest.createHTTPRequest(signer: try client.signer.wait()))
XCTAssertNotNil(request?.headers["Authorization"].first)
```

This saves us an indentation and ensures all lines in the test are always hit.

Ro-M: sounds better this way. I just copied the way another test is written. I'll change that test too.

fabianfett:
Hi @jonnymacs, just verified @Ro-M change works. On `ecs fargate` i tested with the following code:

```swift
import Vapor
import AWSS3

var env = try Environment.detect()
try LoggingSystem.bootstrap(from: &env)
let app = Application(env)
let s3 = S3(region: .eucentral1, httpClientProvider: .shared(app.http.client.shared))

struct Response: Content {
    var contentLength: Int64?
    var contentType: String?
}

app.get("hello") { req -> EventLoopFuture<Response> in
    
    s3.headBucket(.init(bucket: "my.bucket"))
        .flatMap { _ in
            // bucket exists
            s3.headObject(.init(bucket: "my.bucket", key: "03cd03760f29d253c5c689f691aa4343"))
        }
        .map { (output) -> (Response) in
            Response(contentLength: output.contentLength, contentType: output.contentType)
        }
        .flatMapErrorThrowing { (error) -> Response in
            throw error
        }
}

defer { app.shutdown() }
try app.run()
```

Does that void your concerns?

jonnymacs:
@fabianfett is the client using the MetaDataCredentialProvider to obtain credentials?

adam-fowler:
@jonnymacs I've seen this working on ECS as well, with the client using the MetaDataCredentialProvider for credentials. I'm really not sure what your issue was last year when you saw this. Maybe using a signed URL meant some other issue (which has since been fixed) was avoided. Although I can't think what it would be.

jonnymacs:
Thanks guys. I retested my use case and indeed it's working with signed headers. I'll conclude it was fixed in the past year.

sorry for the fire drill!

adam-fowler:
Sorry @Ro-M noticed you didnt get a chance to remove `toHTTPRequestWithSignedURL` before I merged. I'll do it

PR #275, opened 2020-05-25
Move XML stuff to AWSXML and limit exposure therefore

fabianfett:


adam-fowler:
What does @_implementationOnly give us here. Without it people still need to import AWSXML to get access to the XML code. Also I am using the XML code in the streaming response PR. Does this mean we would have two copies of AWSXML.

PR #276, opened 2020-05-25
Provide ServiceConfig on every call

fabianfett:
### Changes

- `AWSClient` does not have the property `serviceConfig` anymore
- the `AWSClient.send` methods have been renamed to `AWSClient.execute` to be more in line with AHC. Further all execute methods now share the same process (private `execute` methods). This should be a huge win for maintainability.
- `AWSClient` is not responsible for creating the `AWSRequest` anymore. `AWSRequest` has now a fail-able initializer. Though the methods are still in the `AWSClient.swift` file for easier code review. Must be moved out after this pr.
- `AWSClient` is not responsible for verifying responses anymore. Ideally this would be implemented as an extension on `AWSHTTPResponse`. For reasons still on `AWSClient` but `verify` is now `static` and can therefore be easily moved out. Next pr.
- All `AWSClientTests` have been moved to `XCTAssertNoThrow` and `XCTAssertThrowsError`

### Open ends

- [ ] `recordMetrics` needs a test
- [ ] `signURL` is unimplemented -> where do we get the `Region` and the `Service` from? Shall we move this method to the Service and it is invoked with a `ServiceConfig` then? @adam-fowler

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -196,26 +196,26 @@ extension AWSClient {
 
         func execute(_ httpRequest: AWSHTTPRequest, attempt: Int) {
             // execute HTTP request
-            _ = httpClient.execute(request: httpRequest, timeout: .seconds(20), on: eventLoop)
+            httpClient.execute(request: httpRequest, timeout: .seconds(20), on: eventLoop)
                 .flatMapThrowing { (response) throws -> Void in
                     // if it returns an HTTP status code outside 2xx then throw an error
                     guard (200..<300).contains(response.status.code) else { throw AWSClient.InternalError.httpResponseError(response) }
                     promise.succeed(response)
                 }
-                .flatMapErrorThrowing { (error)->Void in
-                    // If I get a retry wait time for this error then attempt to retry request
+                .whenFailure { (error) in
+                    // If I we failed, let's try to recover.
+                    // If we get a retry time we will recover after that time has passed
                     if case .retry(let retryTime) = self.retryPolicy.getRetryWaitTime(error: error, attempt: attempt) {
                         // schedule task for retrying AWS request
                         eventloop.scheduleTask(in: retryTime) {
                             execute(httpRequest, attempt: attempt + 1)
                         }
-                    } else if case AWSClient.InternalError.httpResponseError(let response) = error {
-                        // if there was no retry and error was a response status code then attempt to convert to AWS error
-                        promise.fail(self.createError(for: response))
                     } else {
+                        // we could not recover - therefore let's fail
                         promise.fail(error)
                     }
-            }
+                }
+                
         }
 
         execute(httpRequest, attempt: 0)
@@ -273,6 +273,11 @@ extension AWSClient {
             return awsRequest.createHTTPRequest(signer: signer)
         }.flatMap { request in
             return self.invoke(request, on: eventLoop)
+        }.flatMapErrorThrowing { (error) -> AWSHTTPResponse in
+            guard case AWSClient.InternalError.httpResponseError(let response) = error else {
+                throw error
+            }
+            throw self.createError(for: response, configuration: configuration)
         }.map { _ in
             return
         }.recordMetrics(for: configuration.service, operation: operationName)
@@ -310,6 +315,11 @@ extension AWSClient {
             return awsRequest.createHTTPRequest(signer: signer)
         }.flatMap { request in
             return self.invoke(request, on: eventLoop)
+        }.flatMapErrorThrowing { (error) -> AWSHTTPResponse in
+            guard case AWSClient.InternalError.httpResponseError(let response) = error else {
+                throw error
+            }
+            throw self.createError(for: response, configuration: configuration)
         }.map { _ in
             return
         }.recordMetrics(for: configuration.service, operation: operationName)
@@ -342,8 +352,13 @@ extension AWSClient {
             return awsRequest.createHTTPRequest(signer: signer)
         }.flatMap { request in
             return self.invoke(request, on: eventLoop)
+        }.flatMapErrorThrowing { (error) -> AWSHTTPResponse in
+            guard case AWSClient.InternalError.httpResponseError(let response) = error else {
+                throw error
+            }
+            throw self.createError(for: response, configuration: configuration)
         }.flatMapThrowing { response in
-            return try self.validate(operation: operationName, response: response)
+            return try response.validate(operation: operationName, configuration: configuration, middlewares: self.middlewares)
         }.recordMetrics(for: configuration.service, operation: operationName)
     }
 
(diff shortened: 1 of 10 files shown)


adam-fowler:
Can you try not restructuring everything, when making changes.

adam-fowler:
Most of the unrelated changes I'm happy with but I think they need to be in separate PRs. Otherwise when PRs are squashed they get lost.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -234,107 +194,170 @@ extension AWSClient {
...
     ///     - operationName: Name of the AWS operation
     ///     - path: path to append to endpoint URL
     ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
     ///     - input: Input object
+    ///     - configuration: A `ServiceConfig` object configuring the client
+    ///     - eventLoop: An optional `EventLoop` on which the operation shall be send and received
     /// - returns:
     ///     Future containing output object that completes when response is received
-    public func send<Output: AWSDecodableShape, Input: AWSEncodableShape>(operation operationName: String, path: String, httpMethod: String, input: Input, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Output> {
+    public func send<Output: AWSDecodableShape, Input: AWSEncodableShape>(
```
adam-fowler, line 296: execute?

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -234,107 +194,170 @@ extension AWSClient {
...
+            .flatMapErrorThrowing { (error) -> AWSHTTPResponse in
+                // transform the error into an AWSError if possible
+                guard case AWSClient.InternalError.httpResponseError(let response) = error else {
+                    throw error
+                }
+                throw Self.createError(for: response, configuration: configuration)
+            }
+    }
+    
+    private func execute(on eventLoop: EventLoop?, configuration: ServiceConfig, _ createRequest: @escaping () throws -> AWSRequest)
```
adam-fowler, line 342: While the restructure of the send/execute functions ensures we aren't repeating code the code path is a little confusing now. I had to run the code to see how each of the functions were called. Can we add some function headers defining what these functions do individually

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -345,49 +368,55 @@ extension AWSClient {
     /// - returns:
     ///     A signed URL
     public func signURL(url: URL, httpMethod: String, expires: Int = 86400) -> EventLoopFuture<URL> {
-        return signer.map { signer in signer.signURL(url: url, method: HTTPMethod(rawValue: httpMethod), expires: expires) }
```
adam-fowler, line 348: Shouldn't move this to service configuration as you need a signer, which you access from the client. Would be odd to have one code path ie execute go via client and another signURL go via ServiceConfig
```
public func signURL(url: URL, httpMethod: String, configuration: ServiceConfig, expires: Int)
```
I am debating whether to add S3 helper functions in aws-sdk-swift as S3.GetObject and S3.PutObject are the most common uses of this.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -570,17 +606,21 @@ extension AWSClient {
...
 // response validator
-extension AWSClient {
+extension AWSHTTPResponse {
 
     /// Validate the operation response and return a response shape
-    internal func validate<Output: AWSDecodableShape>(operation operationName: String, response: AWSHTTPResponse) throws -> Output {
+    internal func validate<Output: AWSDecodableShape>(
+        operation operationName: String,
+        configuration: ServiceConfig,
+        middlewares: [AWSServiceMiddleware] = []) throws -> Output
```
adam-fowler, line 615: Can we call this clientMiddlewares so it obvious where they are coming from.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -570,17 +606,21 @@ extension AWSClient {
 }
 
 // response validator
-extension AWSClient {
+extension AWSHTTPResponse {
```
adam-fowler, line 609: Request creation has been moved to `AWSRequest` and response validation has been moved to `AWSHTTPResponse`. That's a little unbalanced. Shouldn't most of the Response validation be in AWSResponse?

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -800,14 +843,14 @@ extension AWSClient.ClientError: CustomStringConvertible {
     }
 }
 
-extension AWSClient {
-    func recordMetrics<Output>(_ future: EventLoopFuture<Output>, service: String, operation: String) -> EventLoopFuture<Output> {
+extension EventLoopFuture {
```
adam-fowler, line 846: While extending ELF makes the calling functions slightly cleaner, I do wonder if it makes sense moving AWSClient metrics recording into an unrelated object.

Sources/AWSSDKSwiftCore/Credential/StaticCredential+Shared.swift
```diff
@@ -45,6 +45,7 @@ extension StaticCredential {
         on eventLoop: EventLoop) -> EventLoopFuture<StaticCredential>
     {
         let threadPool = NIOThreadPool(numberOfThreads: 1)
+        threadPool.start()
```
adam-fowler, line 48: Can we move these credential fixes into another PR? I would like them in ASAP, not when the rest of this PR goes in. Did you forget to start the threadPool in your previous PR?

Sources/AWSSDKSwiftCore/Credential/StaticCredential+Shared.swift
```diff
@@ -64,10 +65,11 @@ extension StaticCredential {
         
         return fileIO.openFile(path: filePath, eventLoop: eventLoop)
             .flatMap { (handle, region) in
-                fileIO.read(fileRegion: region, allocator: ByteBufferAllocator(), eventLoop: eventLoop)
+                fileIO.read(fileRegion: region, allocator: ByteBufferAllocator(), eventLoop: eventLoop).map { ($0, handle) }
             }
-            .flatMapThrowing {
-                try Self.sharedCredentials(from: $0, for: profile)
+            .flatMapThrowing { (byteBuffer, handle) in
```
adam-fowler, line 70: If the read throws an error the fileHandle is never closed

Tests/AWSSDKSwiftCoreTests/AWSClientTests.swift
```diff
@@ -78,62 +78,40 @@ class AWSClientTests: XCTestCase {
...
-
-        let awsRequest = try client.createAWSRequest(operation: "test", path: "/", httpMethod: "GET")
+        XCTAssertEqual(config.region, .euwest1)
+        
+        let awsRequest = try AWSRequest(operation: "test", path: "/", httpMethod: "GET", configuration: config)
         XCTAssertEqual(awsRequest.url.absoluteString, "https://service.aws.amazon.com/")
     }
 
+    
     func testCreateAwsRequestWithKeywordInHeader() {
```
adam-fowler, line 101: Is it worth moving all the `AWSRequest` test into a new `AWSRequestTests` given these aren't related to `AWSClient` anymore

adam-fowler:
Closing in favour of #309

PR #295, opened 2020-06-18
Move the AWSClient.validate/createError to AWSResponse

adam-fowler:
Most of `AWSClient.validate` is in `AWSResponse.generateOutputShape`
Most of `AWSClient.createError` is in `AWSResponse.generateError`
Moved all the response validation tests to AWSResponseTests.swift


Sources/AWSSDKSwiftCore/Message/AWSResponse+HAL.swift
```diff
@@ -15,30 +15,28 @@
 import Foundation
```
fabianfett, line 15: Specific import. I know hasn't come up in this pr, but maybe we should do it anyway.

Sources/AWSSDKSwiftCore/Message/AWSResponse.swift
```diff
@@ -78,4 +80,201 @@ public struct AWSResponse {
...
+        }
+        struct RESTJSONError: Codable, ErrorMessage {
+            var code: String?
+            var message: String
+
+            private enum CodingKeys: String, CodingKey {
+                case code = "code"
+                case message = "message"
+            }
+        }
```
fabianfett, line 213: Do we maybe want to move this outside of the function? Would make the whole function easier to read?

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -624,222 +624,32 @@ extension AWSClient {
...
-            }
-            if let error = AWSClientError(errorCode: code, message: errorMessage.message) {
-                return error
-            }
-
-            if let error = AWSServerError(errorCode: code, message: errorMessage.message) {
-                return error
+        // if we can create an AWSResponse and create an error from it return that
+        let awsResponse = try? AWSResponse(from: response, serviceProtocol: serviceConfig.serviceProtocol)
+        if let error = awsResponse?.generateError(serviceConfig: serviceConfig) {
```
fabianfett, line 640:
```suggestion
        if let awsResponse = try? AWSResponse(from: response, serviceProtocol: serviceConfig.serviceProtocol),
            let error = awsResponse?.generateError(serviceConfig: serviceConfig) {
```

PR #296, opened 2020-06-20
Feature/rm refactor awsrequest creation

Ro-M:
## Motivation
This reduces the impact of https://github.com/swift-aws/aws-sdk-swift-core/pull/276

## Changes
- Creation of `AWSRequest` is now moved into an extension on `AWSRequest`
- This essentially impacts 2 overloaded methods on `AWSClient` (`createAWSRequest`) that were previously used to create those `AWSRequests`
- A lot of tests are impacted by this change. Most of the time we now don't need to create a client anymore and can directly use the config
- Moved the `verifyStream` method into an extension of `AWSRequest` and made it static
- Using `XCSkipUnless` to conditionally skip a test to clearly differentiate between succeeding and skipped tests.
- For now I left the new extensions on `AWSRequest` inside the `AWSClient` file to make reading this PR easier. Same goes for the testing methods as a lot of them might now be better suited for a `AWSRequestTests` class.

## Discussion
- In the previously used `createAWSRequest` methods, the `AWSClient` added its middleware by calling `request.applyMiddleware(self.serviceConfig.middlewares + self.middlewares)`. Adding the middlewares now needs to be done manually every time we create an `AWSRequest` in the `AWSClient`. Is this ok? If so, do we need to apply those middlewares in the tests (where this logic was previously implicitly included)? From my point of view it's not needed but I possibly know too little to about the project to have a strong opnion on that ...
- The test `testCreateNIORequest` previously checked if a header with name "x-amz-target" exists. If so, it would assert its value. From my understanding, this request will never contain a header with that name, because the previously used client used a config with `amzTarget` set to `null` (just checked with master: the assertion is indeed ignored because the header does not exist). So I removed this check. Is that ok?


adam-fowler:
>     * Using `XCSkipUnless` to conditionally skip a test to clearly differentiate between succeeding and skipped tests.
> 
`XCSkipUnless` is unavailable in swift 5.1 so you will have to remove these

>     * For now I left the new extensions on `AWSRequest` inside the `AWSClient` file to make reading this PR easier. Same goes for the testing methods as a lot of them might now be better suited for a `AWSRequestTests` class.
> 
I'm quite happy for you to move this out to `AWSRequest.swift` and create a new `AWSRequestTests.swift` for the tests

>     * In the previously used `createAWSRequest` methods, the `AWSClient` added its middleware by calling `request.applyMiddleware(self.serviceConfig.middlewares + self.middlewares)`. Adding the middlewares now needs to be done manually every time we create an `AWSRequest` in the `AWSClient`. Is this ok? If so, do we need to apply those middlewares in the tests (where this logic was previously implicitly included)? From my point of view it's not needed but I possibly know too little to about the project to have a strong opnion on that ...
> 
Yes it is ok that middleware is applied outside of the `init()`. For the tests you don't need to apply middleware if you aren't testing it. Apart from the logging middleware all middleware is service specific.

>     * The test `testCreateNIORequest` previously checked if a header with name "x-amz-target" exists. If so, it would assert its value. From my understanding, this request will never contain a header with that name, because the previously used client used a config with `amzTarget` set to `null` (just checked with master: the assertion is indeed ignored because the header does not exist). So I removed this check. Is that ok?
Ugh that wasn't a great test. Maybe we should add an additional test to verify the x-amz-target is getting through to the request. Create a `ServiceConfig` with `amzTarget` set and verify the x-amz-target header is set with this value.

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
```diff
@@ -138,3 +140,223 @@ public struct AWSRequest {
...
+        
+        self.region = configuration.region
+        self.url = url
+        self.serviceProtocol = configuration.serviceProtocol
+        self.operation = operationName
+        self.httpMethod = httpMethod
+        self.httpHeaders = headers
+        self.body = body
+
+//        .applyMiddlewares(serviceConfig.middlewares + middlewares)
```
adam-fowler, line 330: Remove commented out code.

Tests/AWSSDKSwiftCoreTests/AWSRequestTests.swift
```diff
@@ -0,0 +1,320 @@
...
+            let array: [String]
+        }
+        let input = Input(array: ["entry1", "entry2"])
+        let config = createServiceConfig(serviceProtocol: .ec2, apiVersion: "2013-12-02")
+        var request: AWSRequest?
+        XCTAssertNoThrow(request = try AWSRequest(operation: "Test", path: "/", httpMethod: "GET", input: input, configuration: config))
+        XCTAssertEqual(request?.body.asString(), "Action=Test&Array.1=entry1&Array.2=entry2&Version=2013-12-02")
+    }
+    
+    func testMiddlewareIsOnlyAppliedOnce() throws {
```
adam-fowler, line 300: Don't think the `testMiddlewareIsOnlyAppliedOnce` test is necessary anymore. It was really there to ensure the `AWSClient` wasn't applying the middleware twice. I think we'll need to do a more complex version of this using the `AWSTestServer` to verify this and at that point the test should be in AWSClient.

Remove this and add issue to write new version

Tests/AWSSDKSwiftCoreTests/AWSRequestTests.swift
```diff
@@ -0,0 +1,320 @@
+//
```
adam-fowler, line 1: Need standard file header here, that is why the sanity test is failing

adam-fowler:
This is now merged in. Thanks @Ro-M

PR #301, opened 2020-06-24
AWSRequest uses HTTPHeaders instead of dictionary

adam-fowler:
Also
- remove unused initialiser for `AWSRequest`
- apply content-type and user-agent headers at initialisation
- add support for encoding headers that are dictionaries. Uses pattern set by s3 `x-amz-meta-` headers.

I couldn't do the same for `AWSResponse` as it would have broken the decoding the `x-amz-meta-` headers. Will need to find another way to do this. But I did lowercase all the header names in AWSResponse so we get a consistent casing for response headers.

### DIFF
--- a/Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
+++ b/Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
@@ -18,6 +18,7 @@ import struct Foundation.URL
 import struct Foundation.Date
 import struct Foundation.Data
 import struct Foundation.CharacterSet
+import class Foundation.JSONEncoder
 import struct Foundation.URLComponents
 
 /// Object encapsulating all the information needed to generate a raw HTTP request to AWS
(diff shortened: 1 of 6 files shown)


Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
```diff
@@ -18,6 +18,7 @@ import struct Foundation.URL
 import struct Foundation.Date
 import struct Foundation.Data
 import struct Foundation.CharacterSet
+import class Foundation.JSONEncoder
```
fabianfett, line 21: Why has this been added?

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
```diff
@@ -326,8 +283,26 @@ extension AWSRequest {
...
+                break
+            default:
+                if case .restjson = serviceProtocol, case .raw(_) = body {
+                    httpHeaders.replaceOrAdd(name: "content-type", value: "binary/octet-stream")
+                } else {
+                    httpHeaders.replaceOrAdd(name: "content-type", value: serviceProtocol.contentType)
+                }
+            }
+        }
+        httpHeaders.replaceOrAdd(name: "user-agent", value: "AWSSDKSwift/5.0")
```
fabianfett, line 303:
```suggestion
        httpHeaders.replaceOrAdd(name: "user-agent", value: "AWSSDKSwift/5.0")
        guard httpHeaders["content-type"].first == nil else {
            return
        }
        
        guard httpMethod != "GET", httpMethod != "POST" else {
            return
        }

        if case .restjson = serviceProtocol, case .raw(_) = body {
            httpHeaders.replaceOrAdd(name: "content-type", value: "binary/octet-stream")
        } else {
            httpHeaders.replaceOrAdd(name: "content-type", value: serviceProtocol.contentType)
        }
```

Let's see if we can make this code look a little less complex

adam-fowler:
> I think we should add two tests that check if the standardHeaders have been set, to ensure those aren't removed by accident.

we have `testHeadersAreWritten` already

fabianfett:
> > I think we should add two tests that check if the standardHeaders have been set, to ensure those aren't removed by accident.
> we have testHeadersAreWritten already

sorry. nevermind

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
adam-fowler: Nice typo, in your alternative 
```
guard httpMethod != "GET", httpMethod != "POST" else
```
I think you meant "HEAD"

Sources/AWSSDKSwiftCore/Message/AWSRequest.swift
adam-fowler: removed not sure how that got there

PR #302, opened 2020-06-26
Added CredentialProviderWrapper public interface

adam-fowler:
Given the `CredentialProviders` often need assets from the AWSClient to be constructed properly. I have added a new protocol which is the public interface to `CredentialProviders` called the `CredentialProviderWrapper`. The protocol has one function `getProvider(httpClient:on eventLoop:)`. This returns a `CredentialProvider` given an `AWSHTTPClient` and `EventLoop`. Now everything has a common interface for construction.

- `StaticCredential.getProvider` returns itself
- `EnvironmentCredentialProvider` returns a `StaticCredential`
- `ECSCredentialProvider` returns a RotatingCreditialProvider<ECSMetaDataClient>
- `EC2InstanceCredentialProvider` returns a RotatingCreditialProvider<InstanceMetaDataClient>
- `ConfigFileCredentialProvider` (renamed from shared) returns an internal class

Now that everything has a common interface I have replaced `RuntimeCredentialProvider` with `GroupCredentialProvider` which takes an arbituary list of `Credential Providers` and works through them until it finds one it can use.

You can now create an `AWSClient`  with a single provider as follows
```
let client = AWSClient(credentialProvider: EnvironmentCredentialProvider())
```
or if you want to provide a selection of providers to choose from 
```
let provider = GroupCredentialProvider([
    EnvironmentCredentialProvider(),
    EC2InstanceCredentialProvider(),
    ConfigFileCredentialProvider()
])
let client = AWSClient(credentialProvider: provider)
```
This will attempt to access credentials from the environment, ec2 and then config file. If all of these fail, when someone does a request they will get a `CredentialProviderError.noProviders` error.


adam-fowler:
closed in favour of #303

PR #303, opened 2020-06-26
CredentialProviderFactory with dot notation

fabianfett:
This shall offer an alternative approach to #302.

We might think about wrapping the eventLoop and the httpClient into a Context object, to make breaking changes less likely. @adam-fowler wdyt?

### DIFF
--- a/Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
+++ b/Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
@@ -26,3 +26,85 @@ extension CredentialProvider {
         return
     }
 }
+
+public struct CredentialProviderFactory {
+    
+    public struct Context {
+        let httpClient: AWSHTTPClient
+        let eventLoop: EventLoop
+    }
+    
+    private let cb: (Context) -> CredentialProvider
+    
+    private init(cb: @escaping (Context) -> CredentialProvider) {
+        self.cb = cb
+    }
+    
+    internal func createProvider(context: Context) -> CredentialProvider {
+        self.cb(context)
+    }
+}
+
+extension CredentialProviderFactory {
+    
+    public static func custom(_ factory: @escaping (Context) -> CredentialProvider) -> CredentialProviderFactory {
+        Self(cb: factory)
+    }
+    
+    public static var environment: CredentialProviderFactory {
+        Self() { _ in
+            if let accessKeyId = Environment["AWS_ACCESS_KEY_ID"],
+                let secretAccessKey = Environment["AWS_SECRET_ACCESS_KEY"] {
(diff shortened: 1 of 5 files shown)


adam-fowler:
> We might think about wrapping the eventLoop and the httpClient into a Context object, to make breaking changes less likely. @adam-fowler wdyt?

Yeah go for a context object

adam-fowler:
Can we have an `.empty` as well

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
```diff
@@ -26,3 +26,74 @@ extension CredentialProvider {
...
+    }
+}
+
+extension CredentialProviderFactory {
+    
+    public static func custom(_ factory: @escaping (AWSHTTPClient, EventLoop) -> CredentialProvider) -> CredentialProviderFactory {
+        Self(cb: factory)
+    }
+    
+    public static var environment: CredentialProviderFactory {
```
adam-fowler, line 54: You'll need comments above each as they are public symbols

Sources/AWSSDKSwiftCore/Credential/CredentialProviderError.swift
```diff
@@ -0,0 +1,24 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AWSSDKSwift project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public struct CredentialProviderError: Error, Equatable {
+    enum _CredentialProviderError {
+        case noCredentials
```
adam-fowler, line 17: Looks like `noCredentials` is not used anywhere, my bad

Sources/AWSSDKSwiftCore/Credential/CredentialProvider.swift
```diff
@@ -26,3 +26,85 @@ extension CredentialProvider {
...
+    }
+    
+    public static var runtime: CredentialProviderFactory {
+        Self() { context in
+            RuntimeCredentialProvider.createProvider(
+                on: context.eventLoop,
+                httpClient: context.httpClient)
+        }
+    }
+
```
adam-fowler, line 104: and the custom CredentialProviderFactory version?

adam-fowler: oh and the config file/shared version. I believe this'll need some work. I had to create a new provider for this. Can leave for another PR.

fabianfett: oh and the custom one is in line 55.

PR #304, opened 2020-06-26
Remove `CredentialContainer` protocol, MetaData conforms to ExpiringCredential

fabianfett:


adam-fowler:
Do we need `RotatingCredential` anymore? It is only used in tests now

PR #305, opened 2020-06-26
Added `DeferredCredentialProvider` and added config option

fabianfett:

PR #306, opened 2020-06-27
Runtime Selector Credential Provider

adam-fowler:
Works through a list of CredentialProviderFactories, if it finds one that provides credentials then it uses it. Otherwise it moves onto the next.

Usage is as follows
```
let client = createAWSClient(credentialProvider: .selector(.environment, .ecs, .empty))
```

Eventually this will replace the `RuntimeCredentialProvider`. Just need #305 to be ready


Sources/AWSSDKSwiftCore/Credential/GroupCredentialProvider.swift
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+    /// promise to find a credential provider
+    let startupPromise: EventLoopPromise<CredentialProvider>
+    
+    private let lock = Lock()
+    private var _internalProvider: CredentialProvider? = nil
+
+    init(providers: [CredentialProviderFactory]? = nil, context: CredentialProviderFactory.Context) {
+        #if os(Linux)
```
adam-fowler, line 35: I think the list of providers used is probably something the `AWSClient` should decide, so will make providers list non-optional and move the list construction out to `AWSClient.init`

Tests/AWSSDKSwiftCoreTests/Credential/GroupCredentialProviderTests.swift
```diff
@@ -0,0 +1,208 @@
...
+            switch error {
+            case let error as CredentialProviderError where error == CredentialProviderError.noProvider:
+                break
+            default:
+                XCTFail()
+            }
+        }
+    }
+    
+    /*func testConfigFileProvider() {
```
adam-fowler, line 167: Commented out until #305 is merged

Sources/AWSSDKSwiftCore/Credential/GroupCredentialProvider.swift
adam-fowler: Will now move this to `CredentialProviderFactory.default()` when all providers are ready

PR #309, opened 2020-06-27
Refactor ServiceConfig

Ro-M:
This reduces the impact of #276

## Changes
- rename `ServiceConfig` to `AWSServiceConfig` to prevent symbol clashes with other libraries
- remove `serviceConfig` property from `AWSClient` and the corresponding public accessor properties
- rename all `AWSClient.send` methods to `AWSClient.execute`
- `AWSServiceConfig` must now be provided individually to any of the `AWSClient.execute` methods
- the computed `signer` property on `AWSClient` is now a function (also needs the config)
- adjusted the tests

## Discussion
- the `aws-sdk-swift` package still needs to be adjusted to reflect these changes
- concerning the `AWSClient.createSigner` method: should we really require an `AWSServiceConfig` or is it ok to just provide `signingName` and `region` individually (possibly overloading with a method that still takes the `AWSServiceConfig`)?
- I also needed to add a `serviceConfig` parameter to the `AWSClient.invoke` methods. I opted into naming it `with`. Any better ideas?
- I added documentation for the `serviceConfig` parameter of the `AWSClient.execute` methods. Unsure about if this was necessary or if it is helpful.

### DIFF
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -125,7 +125,7 @@ public final class AWSClient {
 // invoker
 extension AWSClient {
 
-    fileprivate func invoke(_ request: @escaping () -> EventLoopFuture<AWSHTTPResponse>) -> EventLoopFuture<AWSHTTPResponse> {
+    fileprivate func invoke(with serviceConfig: AWSServiceConfig, _ request: @escaping () -> EventLoopFuture<AWSHTTPResponse>) -> EventLoopFuture<AWSHTTPResponse> {
         let eventloop = self.eventLoopGroup.next()
         let promise = eventloop.makePromise(of: AWSHTTPResponse.self)
 
@@ -146,7 +146,7 @@ extension AWSClient {
                         }
                     } else if case AWSClient.InternalError.httpResponseError(let response) = error {
                         // if there was no retry and error was a response status code then attempt to convert to AWS error
-                        promise.fail(self.createError(for: response, serviceConfig: self.serviceConfig))
+                        promise.fail(self.createError(for: response, serviceConfig: serviceConfig))
                     } else {
                         promise.fail(error)
                     }
@@ -159,15 +159,15 @@ extension AWSClient {
     }
 
     /// invoke HTTP request
-    fileprivate func invoke(_ httpRequest: AWSHTTPRequest, on eventLoop: EventLoop) -> EventLoopFuture<AWSHTTPResponse> {
-        return invoke {
+    fileprivate func invoke(_ httpRequest: AWSHTTPRequest, with serviceConfig: AWSServiceConfig, on eventLoop: EventLoop) -> EventLoopFuture<AWSHTTPResponse> {
+        return invoke(with: serviceConfig) {
             return self.httpClient.execute(request: httpRequest, timeout: .seconds(20), on: eventLoop)
         }
     }
 
     /// invoke HTTP request with response streaming
-    fileprivate func invoke(_ httpRequest: AWSHTTPRequest, on eventLoop: EventLoop, stream: @escaping AWSHTTPClient.ResponseStream) -> EventLoopFuture<AWSHTTPResponse> {
-        return invoke {
+    fileprivate func invoke(_ httpRequest: AWSHTTPRequest, with serviceConfig: AWSServiceConfig, on eventLoop: EventLoop, stream: @escaping AWSHTTPClient.ResponseStream) -> EventLoopFuture<AWSHTTPResponse> {
+        return invoke(with: serviceConfig) {
             return self.httpClient.execute(request: httpRequest, timeout: .seconds(20), on: eventLoop, stream: stream)
         }
     }
@@ -209,7 +209,7 @@ extension AWSClient {
                 .applyMiddlewares(serviceConfig.middlewares + self.middlewares)
                 .createHTTPRequest(signer: signer)
         }.flatMap { request in
-            return self.invoke(request, on: eventLoop)
+            return self.invoke(request, with: serviceConfig, on: eventLoop)
         }.map { _ in
             return
         }
@@ -237,7 +237,7 @@ extension AWSClient {
                 .createHTTPRequest(signer: signer)
             
         }.flatMap { request in
-            return self.invoke(request, on: eventLoop)
+            return self.invoke(request, with: serviceConfig, on: eventLoop)
         }.map { _ in
             return
         }
@@ -264,7 +264,7 @@ extension AWSClient {
                 .applyMiddlewares(serviceConfig.middlewares + self.middlewares)
                 .createHTTPRequest(signer: signer)
         }.flatMap { request in
-            return self.invoke(request, on: eventLoop)
+            return self.invoke(request, with: serviceConfig, on: eventLoop)
         }.flatMapThrowing { response in
             return try self.validate(operation: operationName, response: response, serviceConfig: serviceConfig)
         }
@@ -293,7 +293,7 @@ extension AWSClient {
                 .applyMiddlewares(serviceConfig.middlewares + self.middlewares)
                 .createHTTPRequest(signer: signer)
         }.flatMap { request in
-            return self.invoke(request, on: eventLoop)
+            return self.invoke(request, with: serviceConfig, on: eventLoop)
         }.flatMapThrowing { response in
             return try self.validate(operation: operationName, response: response, serviceConfig: serviceConfig)
         }
@@ -314,7 +314,7 @@ extension AWSClient {
                 .applyMiddlewares(serviceConfig.middlewares + self.middlewares)
                 .createHTTPRequest(signer: signer)
         }.flatMap { request in
-            return self.invoke(request, on: eventLoop, stream: stream)
+            return self.invoke(request, with: serviceConfig, on: eventLoop, stream: stream)
         }.flatMapThrowing { response in
             return try self.validate(operation: operationName, response: response, serviceConfig: serviceConfig)
         }
--- a/Sources/AWSSDKSwiftCore/AWSClient.swift
+++ b/Sources/AWSSDKSwiftCore/AWSClient.swift
@@ -52,7 +52,7 @@ public final class AWSClient {
     }
 
     /// AWS service configuration
-    public let serviceConfig: ServiceConfig
+    public let serviceConfig: AWSServiceConfig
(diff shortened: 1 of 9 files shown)


Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -192,143 +181,149 @@ extension AWSClient {
...
     /// - parameters:
     ///     - operationName: Name of the AWS operation
     ///     - path: path to append to endpoint URL
     ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
+    ///     - serviceConfig: configuration to be used to request creation and signing
     ///     - input: Input object
     /// - returns:
     ///     Empty Future that completes when response is received
-    public func send<Input: AWSEncodableShape>(operation operationName: String, path: String, httpMethod: String, input: Input, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Void> {
+    public func execute<Input: AWSEncodableShape>(operation operationName: String, path: String, httpMethod: String, serviceConfig: AWSServiceConfig, input: Input, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Void> {
```
adam-fowler, line 193: Can we break this across multiple lines. It's a very long line eg
```
public func execute<Input: AWSEncodableShape>(
    operation operationName: String,
    path: String,
    ...
) -> EventLoopFuture<Void>
```

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -192,143 +181,149 @@ extension AWSClient {
...
+    /// execute an empty request and return a future with an empty response
     /// - parameters:
     ///     - operationName: Name of the AWS operation
     ///     - path: path to append to endpoint URL
     ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
+    ///     - serviceConfig: configuration to be used to request creation and signing
     /// - returns:
     ///     Empty Future that completes when response is received
-    public func send(operation operationName: String, path: String, httpMethod: String, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Void> {
+    public func execute(operation operationName: String, path: String, httpMethod: String, serviceConfig: AWSServiceConfig, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Void> {
```
adam-fowler, line 222: breakup across multiple lines as above

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -192,143 +181,149 @@ extension AWSClient {
...
     /// - parameters:
     ///     - operationName: Name of the AWS operation
     ///     - path: path to append to endpoint URL
     ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
+    ///     - serviceConfig: configuration to be used to request creation and signing
     /// - returns:
     ///     Future containing output object that completes when response is received
-    public func send<Output: AWSDecodableShape>(operation operationName: String, path: String, httpMethod: String, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Output> {
+    public func execute<Output: AWSDecodableShape>(operation operationName: String, path: String, httpMethod: String, serviceConfig: AWSServiceConfig, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Output> {
         let eventLoop = eventLoop ?? eventLoopGroup.next()
```
adam-fowler, line 252: breakup across multiple lines as above

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -192,143 +181,149 @@ extension AWSClient {
...
     ///     - operationName: Name of the AWS operation
     ///     - path: path to append to endpoint URL
     ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
+    ///     - serviceConfig: configuration to be used to request creation and signing
     ///     - input: Input object
     /// - returns:
     ///     Future containing output object that completes when response is received
-    public func send<Output: AWSDecodableShape, Input: AWSEncodableShape>(operation operationName: String, path: String, httpMethod: String, input: Input, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Output> {
+    public func execute<Output: AWSDecodableShape, Input: AWSEncodableShape>(operation operationName: String, path: String, httpMethod: String, serviceConfig: AWSServiceConfig, input: Input, on eventLoop: EventLoop? = nil) -> EventLoopFuture<Output> {
         let eventLoop = eventLoop ?? eventLoopGroup.next()
```
adam-fowler, line 281: breakup across multiple lines as above

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -192,143 +181,149 @@ extension AWSClient {
...
+    /// execute a request with an input object and return a future with the output object generated from the response
+    /// - parameters:
+    ///     - operationName: Name of the AWS operation
+    ///     - path: path to append to endpoint URL
+    ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
+    ///     - serviceConfig: configuration to be used to request creation and signing
+    ///     - input: Input object
+    /// - returns:
+    ///     Future containing output object that completes when response is received
+    public func execute<Output: AWSDecodableShape, Input: AWSEncodableShape>(operation operationName: String, path: String, httpMethod: String, serviceConfig: AWSServiceConfig, input: Input, on eventLoop: EventLoop? = nil, stream: @escaping AWSHTTPClient.ResponseStream) -> EventLoopFuture<Output> {
```
adam-fowler, line 310: breakup across multiple lines as above

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -337,29 +332,27 @@ extension AWSClient {
     ///     - url : URL to sign
     ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
     ///     - expires: How long before the signed URL expires
+    ///     - serviceConfig: Additional configuration to sign the url
     /// - returns:
     ///     A signed URL
-    public func signURL(url: URL, httpMethod: String, expires: Int = 86400) -> EventLoopFuture<URL> {
-        return signer.map { signer in signer.signURL(url: url, method: HTTPMethod(rawValue: httpMethod), expires: expires) }
+    public func signURL(url: URL, httpMethod: String, expires: Int = 86400, serviceConfig: AWSServiceConfig) -> EventLoopFuture<URL> {
```
adam-fowler, line 338: Let's stick with serviceConfig here. The signing name is not always obvious for each service. 
```
return client.signURL(url: myURL, httpMethod: "GET", serviceConfig: S3.serviceConfig)
```

Tests/AWSSDKSwiftCoreTests/PaginateTests.swift
```diff
@@ -26,13 +26,15 @@ class PaginateTests: XCTestCase {
     var eventLoopGroup: EventLoopGroup!
     var httpClient: AWSHTTPClient!
     var client: AWSClient!
+    var config: AWSServiceConfig!
```
adam-fowler, line 29: no need to wait until the `setUp` function to create this. Although this is fine as well. leave it if you want, as you like.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -192,143 +181,149 @@ extension AWSClient {
 // public facing apis
 extension AWSClient {
 
-    /// send a request with an input object and return a future with an empty response
+    /// execute a request with an input object and return a future with an empty response
     /// - parameters:
     ///     - operationName: Name of the AWS operation
     ///     - path: path to append to endpoint URL
     ///     - httpMethod: HTTP method to use ("GET", "PUT", "PUSH" etc)
+    ///     - serviceConfig: configuration to be used to request creation and signing
```
adam-fowler, line 189: Maybe "AWS service configuration used in request creation and signing". Ha just noticed I hadn't put comments about the `eventLoop` and `stream` parameters in those functions. I'll do that once this is in.

Sources/AWSSDKSwiftCore/AWSClient.swift
```diff
@@ -130,7 +119,7 @@ public final class AWSClient {
 // invoker
 extension AWSClient {
 
-    fileprivate func invoke(_ request: @escaping () -> EventLoopFuture<AWSHTTPResponse>) -> EventLoopFuture<AWSHTTPResponse> {
+    fileprivate func invoke(with serviceConfig: AWSServiceConfig, _ request: @escaping () -> EventLoopFuture<AWSHTTPResponse>) -> EventLoopFuture<AWSHTTPResponse> {
```
adam-fowler, line 122: `with serviceConfig` is fine

Sources/AWSSDKSwiftCore/AWSClient.swift
Ro-M: sure we can. do you just want this specific part to be split to multiple lines or should I also adjust the other method (any hard-ish character limit per line?)

Tests/AWSSDKSwiftCoreTests/PaginateTests.swift
Ro-M: my intention was to put the setup code all in one place together. but you are right, we can assign it directly at the declaration, maybe also make it a let. I'll see

Sources/AWSSDKSwiftCore/AWSClient.swift
Ro-M: yeah that sounds alright, specifying that it is a config for AWS

Tests/AWSSDKSwiftCoreTests/PaginateTests.swift
Ro-M: the service config uses the address of the server which is initialized in `setup`. so I would just leave the creation of config in the setup as well. other options would be to make it a lazy var but I think the way it is now is more readable.

Sources/AWSSDKSwiftCore/AWSClient.swift
adam-fowler: We will be using swiftformat in the near future so these will be done automatically, but worthwhile sorting now

Tests/AWSSDKSwiftCoreTests/PaginateTests.swift
adam-fowler: fair point

PR #313, opened 2020-06-29
Deferred credential provider version 2

adam-fowler:
@fabianfett This is an alternative to your version that takes a closure. I have tried to keep all the code in the same place so they are easy to compare. 

Setup interface to be the same as RotatingCredentialProvider. ie `init(eventLoop: EventLoop, client: CredentialProvider)`
Moved the StaticCredential+Shared extension into a `ConfigFileCredentialProvider` struct.
Fixed up tests
Added test to verify the client credential provider code is only called once.


adam-fowler:
In theory we could just use `RotatingCredentialProvider`. The `DeferredCredentialProvider` doesn't need to check if credentials are expiring. A minor optimisation

PR #350, opened 2020-08-30
Body.json now holds a ByteBuffer instead of Data

adam-fowler:


Sources/AWSSDKSwiftCore/Message/AWSResponse+HAL.swift
```diff
@@ -17,31 +17,36 @@ import NIO
 
 // AWS HAL services I know of are APIGateway, Pinpoint, Greengrass
 extension AWSResponse {
-    /// process hal+json date. Extract properties from HAL
-    func getHypertextApplicationLanguageBody() throws -> Body {
-        guard case .json(let buffer) = self.body,
+    /// return if body is hypertext application language
+    /// - Returns: <#description#>
+    func IsHypertextApplicationLanguage() -> Bool {
```
fabianfett, line 22: This is more a readonly `var`, isn't it?

adam-fowler: Ok can change.

adam-fowler: done

fabianfett (approved):

PR #360, opened 2020-09-13
ConfigFileCredentialProvider resolves tilde to posix home directory instead of sandboxed filesyste,

sebsto:
This PR changes `Sources/AWSSDKSwiftCore/Credential/ConfigFileCredentialProvider.swift` to fix https://github.com/swift-aws/aws-sdk-swift-core/issues/359
I updated the corresponding test `Tests/AWSSDKSwiftCoreTests/Credential/ConfigFileCredentialProviderTests.swift`

Two other commits are part of this PR 
- `sanity.sh` : I added a line to verify if `swiftformat` is installed or not and provide installation instruction 
- I added two very simple `build.sh`and `test.sh`to help newcomers to get started

### DIFF
--- a/Sources/AWSSDKSwiftCore/Credential/ConfigFileCredentialProvider.swift
+++ b/Sources/AWSSDKSwiftCore/Credential/ConfigFileCredentialProvider.swift
@@ -19,7 +19,11 @@ import NIO
 #if os(Linux)
 import Glibc
 #else
+#if os(macOS)
+import Foundation
+#else
 import Foundation.NSString
+#endif
 #endif
 
 struct AWSConfigFileCredentialProvider: CredentialProvider {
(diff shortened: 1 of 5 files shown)


adam-fowler (commented):
Thank you, for the PR. I think we can do this in a simpler manner though. We already have a solution for Linux, can we not use that solution for macOS. See comments in code

Also can we remove script/build.sh and script/test.sh. I don't think they really help newcomers. I would expect newcomers already know how to build swift projects and script/test.sh makes assumptions about what needs to be tested. Plus script/test.sh isn't doing what I'd expect it to. I will add a comment at the top of the docker-compose file indicating how I expect it to be run.

sebsto:
I somehow manage to screw up my repository after the project name change. (your branch is 807 commits ahead of soto/soto-core). So I will close this PR and send a fresh one. Apologies for the churn

Issue #359, opened 2020-09-12
AWSSwiftSDKCore can not locate config file in sandboxed environments 


sebsto:
**Describe the bug**
When using ConfigFileCredentialProvider on MacOS in a Sandboxed app, the path `~/.aws/credentials` is expanded to `/Users/name/Library/Containers/bundle-id/Data/.aws/credentials`

The `NSString.expandingTildeInPath` here 
https://github.com/swift-aws/aws-sdk-swift-core/blob/main/Sources/AWSSDKSwiftCore/Credential/ConfigFileCredentialProvider.swift#L137
must be replaced with posix code such as 

```swift 
    let pw = getpwuid(getuid())
    let home = pw?.pointee.pw_dir
```

**To Reproduce**
Create a sandboxed app, initialize the AWS SDK with 

```swift 
        self.awsClient = AWSClient(credentialProvider: .configFile(), middlewares: [AWSLoggingMiddleware()], httpClientProvider: .createNew)
        self.s3 = S3(client: awsClient)
```

The code produce the error message `No credential provider found` at line https://github.com/swift-aws/aws-sdk-swift-core/blob/main/Sources/AWSSDKSwiftCore/Credential/DeferredCredentialProvider.swift#L49

**Expected behavior**
credentials file should be loaded

**Setup (please complete the following information):**
 - OS: [e.g. iOS] Mac OS 10.15
 - Version of aws-sdk-swift [e.g. 3.0] 5.x.x branch
 - Authentication mechanism [hard-coded credentials, IAM Instance Profile on EC2, etc]  ConfigFile

**Additional context**

adam-fowler:
@sebsto have you verified this code. Surely a sandboxed app would not have access to the ~/.aws/credential file.

sebsto:
Yes, it works, providing the app has the correct entitlements.  Reading on ~/.aws requires explicit user selection (through a `NSOpenPanel()`) I also tested `~/Downloads`

I am sending a PR which is harmless for non sandboxed environments and solves this for sandboxed environments

PR #361, opened 2020-09-14
macOS tilde expansion in path for sandboxed app

sebsto:
As discussed, here is a new PR to address https://github.com/soto-project/soto-core/issues/359

The proposed solution to handle Linux and macOS by just changing [this line](https://github.com/soto-project/soto-core/blob/main/Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift#L116) does not work, because when running in a sandboxed app `wexp.we_wordv` is nil.

I revert back to my original solution. Tested on macOS Catalina on both a sanboxed (with appropriate entitlements for file access) and not sandboxed app.

### DIFF
--- a/Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
+++ b/Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
@@ -140,7 +140,7 @@ struct AWSConfigFileCredentialProvider: CredentialProvider {
         else {
             return filePath
         }
-        return filePath.replacingOccurrences(of: "~", with: homePath)
+        return filePath.starts(with: "~") ? homePath + filePath.dropFirst() : filePath
         #else
         return NSString(string: filePath).expandingTildeInPath
         #endif
--- a/Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
+++ b/Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
@@ -133,17 +133,16 @@ struct AWSConfigFileCredentialProvider: CredentialProvider {
 
             return pth
         }
-        #else
-        #if os(macOS)
-        guard let home = getpwuid(getuid())?.pointee.pw_dir else {
+        #elseif os(macOS)
+        // can not use wordexp on macOS because for sandboxed application wexp.we_wordv == nil
+        guard let home = getpwuid(getuid())?.pointee.pw_dir,
+            let homePath = String(cString: home, encoding: .utf8)
+        else {
             return filePath
         }
-        let homePath = FileManager.default.string(withFileSystemRepresentation: home, length: Int(strlen(home)))
-        let expandedPath = filePath.replacingOccurrences(of: "~", with: homePath)
-        return expandedPath
+        return filePath.replacingOccurrences(of: "~", with: homePath)
         #else
         return NSString(string: filePath).expandingTildeInPath
-        #endif
         #endif
     }
 }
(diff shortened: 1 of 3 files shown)


Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
```diff
@@ -134,7 +134,16 @@ struct AWSConfigFileCredentialProvider: CredentialProvider {
             return pth
         }
         #else
+        #if os(macOS)
+        guard let home = getpwuid(getuid())?.pointee.pw_dir else {
+            return filePath
+        }
+        let homePath = FileManager.default.string(withFileSystemRepresentation: home, length: Int(strlen(home)))
+        let expandedPath = filePath.replacingOccurrences(of: "~", with: homePath)
```
adam-fowler, line 142: Any reason we can't use `String(cString: home)` instead of `FileManager.default.string(withFileSystemRepresentation:)` here?

Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
```diff
@@ -134,7 +134,16 @@ struct AWSConfigFileCredentialProvider: CredentialProvider {
             return pth
         }
         #else
+        #if os(macOS)
+        guard let home = getpwuid(getuid())?.pointee.pw_dir else {
```
adam-fowler, line 138: Can we add a comment indicating why we cannot use `wordexp` here

Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
```diff
@@ -134,7 +134,16 @@ struct AWSConfigFileCredentialProvider: CredentialProvider {
             return pth
         }
         #else
+        #if os(macOS)
```
adam-fowler, line 137: Can we make this `#elseif os(macOS)`

Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
sebsto: no reason, ignorance on my side :-) 
What's the reason your pushing to not use FileManager ? Performance ? Binaries size ?

sebsto: change is done

adam-fowler: Limiting our exposure to Foundation, although in this situation it isn't such an issue as this is only running on macOS.

Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
```diff
@@ -133,6 +133,14 @@ struct AWSConfigFileCredentialProvider: CredentialProvider {
...
             return pth
         }
+        #elseif os(macOS)
+        // can not use wordexp on macOS because for sandboxed application wexp.we_wordv == nil
+        guard let home = getpwuid(getuid())?.pointee.pw_dir,
+            let homePath = String(cString: home, encoding: .utf8)
+        else {
+            return filePath
+        }
+        return filePath.replacingOccurrences(of: "~", with: homePath)
```
adam-fowler, line 143: Should this only be replaced if it is the first character of the path?

sebsto: although it is very unlikely to have other tildes in such path, you're right !

adam-fowler: It's better we are correct. Look up bash tilda expansion and it gets all complex. Let's go for checking as a prefix

sebsto: yep, I just changed it

adam-fowler (approved):

PR #371, opened 2020-09-20
Removes duplicate "/" in InstanceMetaDataServiceProvider request

t089:
When the InstanceMetaDataServiceProvider requests credentials for
the instance role, the URI contained two subsequent "//" slashes.
This resulted in a `MOVED PERMANENTLY` redirect which was not
followed, and thus failing the request.

### DIFF
--- a/Sources/AWSSDKSwiftCore/MetaDataService.swift
+++ b/Sources/AWSSDKSwiftCore/MetaDataService.swift
@@ -204,7 +204,7 @@ struct InstanceMetaDataServiceProvider: MetaDataServiceProvider {
                 let roleName = String(data: response.body, encoding: .utf8) else {
                     throw MetaDataServiceError.couldNotGetInstanceRoleName
             }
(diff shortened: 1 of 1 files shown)


adam-fowler:
@t089 Yeah this looks good. Out of interest when you had the issue were you running inside a docker container?

adam-fowler (approved):

t089:
Yup, on k8s (EKS). Btw I think it would make sense to enable this MetadataService even on macOS. I often use `aws-vault` which simulates an EC2 metadata service running on your Mac to provide credentials and was surprised to find it does not work with this SDK on macOS.

adam-fowler:
Credential acquisition has been completely rewritten for v5. It is a lot more flexible. You can choose which credential providers you run, even write your own, and therefore could choose to run the ec2 instance metadata credential provider on your Mac.

adam-fowler:
This is now in patch release 4.7.1

PR #372, opened 2020-09-21
Replace logger with baggage context for AWSClient

adam-fowler:
Changes required to the public API of Soto for integrating with the swift tracing code. It is not intended that I add instrumentation code here. All we are doing is passing in the `Context`

The PR has a fairly limited scope in that it only covers the interface functions of AWSClient and passes the `Context` down to the HTTP client. Although we can't actually pass it to `AsyncHTTPClient.HTTPClient` yet. 

I have another branch which also passes the `Context` to the `CredentialProviders` but didn't include it here as the number of files that were touched by that change would make this difficult to review.

### DIFF
--- a/Sources/SotoCore/AWSClient.swift
+++ b/Sources/SotoCore/AWSClient.swift
@@ -486,7 +486,7 @@ extension AWSClient {
         httpMethod: HTTPMethod,
         headers: HTTPHeaders = HTTPHeaders(),
         expires: TimeAmount,
-        serviceConfig: AWSServiceConfig,
+        config: AWSServiceConfig,
         context: Context = AWSClient.defaultBaggageContext()
     ) -> EventLoopFuture<URL> {
         var context = context
--- a/Sources/SotoCore/AWSClient.swift
+++ b/Sources/SotoCore/AWSClient.swift
@@ -259,7 +259,7 @@ extension AWSClient {
                 )
             },
             execute: { request, eventLoop, logger in
-                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, logger: logger)
+                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, context: context)
             },
             processResponse: { _ in
                 return
@@ -299,7 +299,7 @@ extension AWSClient {
                 )
             },
             execute: { request, eventLoop, logger in
-                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, logger: logger)
+                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, context: context)
             },
             processResponse: { _ in
                 return
@@ -339,7 +339,7 @@ extension AWSClient {
                 )
             },
             execute: { request, eventLoop, logger in
-                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, logger: logger)
+                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, context: context)
             },
             processResponse: { response in
                 return try self.validate(operation: operationName, response: response, serviceConfig: serviceConfig)
@@ -382,7 +382,7 @@ extension AWSClient {
                 )
             },
             execute: { request, eventLoop, logger in
-                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, logger: logger)
+                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, context: context)
             },
             processResponse: { response in
                 return try self.validate(operation: operationName, response: response, serviceConfig: serviceConfig)
@@ -427,7 +427,7 @@ extension AWSClient {
                 )
             },
             execute: { request, eventLoop, logger in
-                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, logger: logger, stream: stream)
+                return self.httpClient.execute(request: request, timeout: serviceConfig.timeout, on: eventLoop, context: context, stream: stream)
             },
             processResponse: { response in
                 return try self.validate(operation: operationName, response: response, serviceConfig: serviceConfig)
--- a/Sources/SotoCore/AWSClient.swift
+++ b/Sources/SotoCore/AWSClient.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import AsyncHTTPClient
+import BaggageContext
 import Dispatch
 import struct Foundation.URL
 import struct Foundation.URLQueryItem
@@ -20,8 +21,6 @@ import Logging
 import Metrics
 import NIO
 import NIOConcurrencyHelpers
-import NIOHTTP1
-import NIOTransportServices
 import SotoSignerV4
 import SotoXML
 
@@ -62,6 +61,8 @@ public final class AWSClient {
 
     /// default logger that logs nothing
     public static let loggingDisabled = Logger(label: "AWS-do-not-log", factory: { _ in SwiftLogNoOpLogHandler() })
+    /// default baggage context
+    public static let emptyContext = DefaultContext.TODO(logger: loggingDisabled)
 
     static let globalRequestID = NIOAtomic<Int>.makeAtomic(value: 0)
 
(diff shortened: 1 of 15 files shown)


Sources/SotoCore/AWSClient.swift
```diff
@@ -62,6 +61,8 @@ public final class AWSClient {
 
     /// default logger that logs nothing
     public static let loggingDisabled = Logger(label: "AWS-do-not-log", factory: { _ in SwiftLogNoOpLogHandler() })
+    /// default baggage context
+    public static let emptyContext = DefaultContext.TODO(logger: loggingDisabled)
```
adam-fowler, line 65: This used as the default context when no context is provided

Sources/SotoCore/AWSClient.swift
```diff
@@ -419,44 +426,47 @@ extension AWSClient {
...
-        logger: Logger = AWSClient.loggingDisabled
+        context: Context,
+        on eventLoop: EventLoop? = nil
     ) -> EventLoopFuture<Output> {
         let eventLoop = eventLoop ?? eventLoopGroup.next()
-        let logger = logger.attachingRequestId(Self.globalRequestID.add(1), operation: operationName, service: config.service)
+        var context = context
+        context.baggage.awsService = config.service
+        context.baggage.awsOperation = operationName
+        context.baggage.awsRequestId = Self.globalRequestID.add(1)
```
adam-fowler, line 455: Have to make a copy of the context here to add values to the baggage

Sources/SotoCore/Baggage.swift
```diff
@@ -0,0 +1,60 @@
...
+    var awsRequestId: Int? {
+        get {
+            self[_key: RequestIdKey.self]
+        }
+        set {
+            self[_key: RequestIdKey.self] = newValue
+        }
+
+    }
+}
```
adam-fowler, line 60: Is it preferred to have one structure that holds all the data related to one system or individual keys for each element. The individual keys produces Logger output that is easier to process.

adam-fowler:
@ktoso @pokryfka @fabianfett

Sources/SotoCore/AWSClient.swift
```diff
@@ -419,44 +426,47 @@ extension AWSClient {
...
+            on: eventLoop
         )
     }
 
     /// internal version of execute
     internal func execute<Output>(
         operation operationName: String,
         createRequest: @escaping () throws -> AWSRequest,
-        execute: @escaping (AWSHTTPRequest, EventLoop, Logger) -> EventLoopFuture<AWSHTTPResponse>,
+        execute: @escaping (AWSHTTPRequest, EventLoop, Context) -> EventLoopFuture<AWSHTTPResponse>,
```
pokryfka, line 445: I think its private API, still it would be nice to keep the arguments in the same order for consistency (Context with optional EventLoop last)?

adam-fowler: The EventLoop isn't optional in this case but fair enough

Sources/SotoCore/AWSClient.swift
```diff
@@ -466,26 +476,30 @@ extension AWSClient {
...
     public func signURL(
         url: URL,
         httpMethod: HTTPMethod,
         headers: HTTPHeaders = HTTPHeaders(),
-        expires: Int = 86400,
-        serviceConfig: AWSServiceConfig,
-        logger: Logger = AWSClient.loggingDisabled
+        config: AWSServiceConfig,
+        context: Context = AWSClient.emptyContext,
+        expires: Int = 86400
```
pokryfka, line 488: to be honest I find it a bit strange that expires has a default value, is it for consistency with other (official) AWS frameworks?

most API has "input" before the context, so the signature is a bit inconsistent here

pokryfka:
Awesome, this will make adding tracing much easier.

Will that be included in 5.0.0 or after?

adam-fowler:
> Will that be included in 5.0.0 or after?

Yes I would like to get this in 5.0.0. Depends on when we get an official swift repo.

ktoso:
Awesome, thanks for the ping! I’m offline one more day (national holidays here, will get to it the day after), looking forward to see it in a real repo like this. Hope not many warts remain. Thank you for pushing forward here!

Sources/SotoCore/AWSClient.swift
adam-fowler: Yeah you are probably correct here. I think I added the default in the signer and it just bubbled through all the APIs. I'm gonna make this a `TimeAmount` as well. Makes for easier reading, but will do in separate PR though

adam-fowler: See #373

Sources/SotoCore/Baggage.swift
ktoso: individual is fine 👍 
Good that you're using the `var name` (well, now it'll be `var nameOverride`...), since that helps the logger to print them nicely 👍

Sources/SotoCore/Exports.swift
```diff
@@ -15,6 +15,8 @@
 @_exported import protocol SotoSignerV4.Credential
 @_exported import struct SotoSignerV4.StaticCredential
 
+@_exported import protocol BaggageContext.Context
```
ktoso, line 18: Is this export necessary? People only stumble into the `Context` as a parameter to Soco functions right? I don't think they need to use it explicitly?

Sources/SotoTestUtils/TestUtils.swift
```diff
@@ -113,4 +114,8 @@ public struct TestEnvironment {
         }
         return AWSClient.loggingDisabled
     }()
+
+    public static var context: Context = {
+        DefaultContext.topLevel(logger: logger)
+    }()
```
ktoso, line 120: Ah for testing only right? Okey. Would be worried if this was for the real sources, but in testing okey 👍

ktoso (approved):
Looks good, thanks a lot for taking a stab at it. :)

It seems some names may still change due to the ongoing reviews... 
FYI: I think expecting the 25th be the deadline when names are set in stone is reasonable 👍

Sources/SotoCore/AWSClient.swift
adam-fowler: I am keen to keep a default context as this library isn't only used on server applications. It is also used in macOS and iOS client applications and I don't think the tracing context is as useful in those situations. Also I would suggest Soto is less glue than something like the http client.

I'm happy to change the `let` to the suggested `func`.
I'll have a think about the naming, not completely with `.todo` here given the possiblity of this being used as a client lib.

Sources/SotoCore/AWSClient.swift
adam-fowler: That'd be slightly cleaner, we did discuss this here https://github.com/slashmo/gsoc-swift-baggage-context/pull/34#discussion_r488048198 although I can't find your response, in that situation my suggestion was to return a newly constructed `DefaultContext` which you rightly pointed out would cause confusion. But now that baggage is a `var` something of this form should be possible.
```
extension Context {
    func withBaggage(_: (inout Baggage) -> ()) -> Self
}
```

Sources/SotoCore/Exports.swift
adam-fowler: Fair enough. If they aren't creating their then this is unnecessary

Sources/SotoTestUtils/TestUtils.swift
adam-fowler: Just for testing

adam-fowler:
> Looks good, thanks a lot for taking a stab at it. :)
> 
> It seems some names may still change due to the ongoing reviews...
> FYI: I think expecting the 25th be the deadline when names are set in stone is reasonable 👍

Good luck with fixing the names. A lot of people are pulling in different directions

Sources/SotoCore/AWSClient.swift
ktoso: I see, thanks for the reminder. I guess you might want to use `.topLevel` then after all. 
It's "more okey" for this lib than httpclient to provide a default here, feel free to stick with the default 👍

adam-fowler: Updated `emptyContext` to function, renamed to `defaultBaggageContext`. I'm going to keep it as a TODO because people will want to catch when they aren't passing context to soto.

Sources/SotoCore/Exports.swift
adam-fowler: removed, along with the Logging export

adam-fowler:
Closed in favour of #396

PR #396, opened 2020-11-01
Replace logger with baggage context for AWSClient v2

adam-fowler:
This is version 2 of placing the logger parameters with a baggage context. Merging in latest changes was too much of a pain. Original PR can be found here #372 

Changes required to the public API of Soto for integrating with the swift tracing code. It is not intended that I add instrumentation code here. All we are doing is passing in the Context

The PR has a fairly limited scope in that it only covers the interface functions of AWSClient and passes the Context down to the HTTP client. Although we can't actually pass it to AsyncHTTPClient.HTTPClient yet.


adam-fowler:
closed in favour of #443

PR #397, opened 2020-11-01
Allows to create an existing Region from a String

sebsto:
as discussed on Slack


Sources/SotoCore/Doc/Region.swift
```diff
@@ -80,8 +80,8 @@ public struct Region: RawRepresentable, Equatable {
     public static func other(_ name: String) -> Region { .init(rawValue: name) }
 }
 
-extension Region {
-    public var partition: AWSPartition {
+public extension Region {
+    var partition: AWSPartition {
```
adam-fowler, line 84: Any reason you changed this?

Sources/SotoCore/Doc/Region.swift
```diff
@@ -156,8 +156,8 @@ public struct AWSPartition: RawRepresentable, Equatable, Hashable {
     public static var awsisob: AWSPartition { .init(partition: .awsisob) }
 }
 
-extension AWSPartition {
-    public var dnsSuffix: String {
+public extension AWSPartition {
+    var dnsSuffix: String {
```
adam-fowler, line 160: Any reason you changed this?

Sources/SotoCore/Doc/Region.swift
```diff
@@ -167,3 +167,70 @@ extension AWSPartition {
         }
     }
 }
+
+// allows to create a Region from a String
+// this initialializer returns an Optional.
+// it is non nil for valid region name only
+public extension Region {
+    init?(regionName: String) {
```
adam-fowler, line 175: Can we change the parameter name to `awsRegionName`. This initializer is only needed if you want to verify the region is an AWS region

Sources/SotoCore/Doc/Region.swift
```diff
@@ -167,3 +167,70 @@ extension AWSPartition {
         }
     }
 }
+
+// allows to create a Region from a String
+// this initialializer returns an Optional.
+// it is non nil for valid region name only
+public extension Region {
```
adam-fowler, line 174: Can you change the comment to indicate this will only create a Region if the provided region name is a valid AWS region name. We already have an initializer that allows you create a Region from a String.

Sources/SotoCore/Doc/Region.swift
```diff
@@ -167,3 +167,70 @@ extension AWSPartition {
...
+    init?(regionName: String) {
+        switch regionName {
+        case "af-south-1":
+            self = Region.afsouth1
```
adam-fowler, line 178: Can we change this to
```swift
        self.init(rawValue: regionName)
        switch self {
        case .euwest1,
             .euwest2,
                 ...
             return
        default:
            return nil

...
```

Tests/SotoCoreTests/Doc/RegionFromStringTests.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import SotoCore
+import XCTest
+
+class StringRegionTests: XCTestCase {
+    func testStringToRegion() {
```
adam-fowler, line 19: There is a `testRegionEnum` in `AWSClientTests` can you move that function here

Sources/SotoCore/Doc/Region.swift
sebsto: swiftformat did

Sources/SotoCore/Doc/Region.swift
sebsto: The problem with this is Stencil. The last line finishes with `.euwest2 :` (colon and not coma)
Is there a way to tell Stencil to loop on all occurences, except the last one ?

Sources/SotoCore/Doc/Region.swift
adam-fowler: Have you upgraded to swift format v0.47? Soto is still using v0.46.3.

Sources/SotoCore/Doc/Region.swift
adam-fowler: You can use `{%if not forloop.last %}`

Sources/SotoCore/Doc/Region.swift
sebsto: I did, it was part of a a `brew upgrade`. would you recommend to stick to 0.46.3 ?

Sources/SotoCore/Doc/Region.swift
sebsto: swiftformat....

Sources/SotoCore/Doc/Region.swift
adam-fowler: There are differences between the output of 0.46.3 and 0.47. You might have issue submitting code, if the outputs disagree. I'll upgrade in the near future, but for the moment I'm sticking to 0.46.3

Sources/SotoCore/Doc/Region.swift
sebsto: done

Sources/SotoCore/Doc/Region.swift
sebsto: done

Sources/SotoCore/Doc/Region.swift
sebsto: done

Tests/SotoCoreTests/Doc/RegionFromStringTests.swift
sebsto: done. I am considering generating the RegionTest file automatically too, to improve test coverage

adam-fowler:
@sebsto I can see what you're trying to do with the automatically generating the tests, but given you are using the same method to generate the tests if there was an issue with the generation of the regions, I imagine this same issue would occur with the tests. Then possibly it would not catch the issue.

sebsto:
> @sebsto I can see what you're trying to do with the automatically generating the tests, but given you are using the same method to generate the tests if there was an issue with the generation of the regions, I imagine this same issue would occur with the tests. Then possibly it would not catch the issue.

I am well aware of that risk. What would you suggest ? I can rewrite the code that generate the region test (using SSM API for example), it would not be that extensive (we will miss the two ISO regions and China) but that would give an independent test of the Region class.

sebsto:
Sorry for the delay, I have been busy at work. I rewrote the `generate-region-test.swift` script to generate the test file based on the region list provided by SSM `getParametersByPath`

adam-fowler (approved):

PR #399, opened 2020-11-03
Logger Eventloop parameter ordering

adam-fowler:
Re-order Logger and EventLoop parameters in public functions of AWSClient so it is easier to replace Logger with BaggageContext at a later date


ktoso (approved):
This looks good, thank you.

The rule indeed is "context as last non-optional parameter (before any function parameters)" so this looks good in and aligns to that 👍

adam-fowler:
> This looks good, thank you.
> 
> The rule indeed is "context as last non-optional parameter (before any function parameters)" so this looks good in and aligns to that 👍

Ha! @ktoso didn't expect that. I only posted a few minutes back. You got some notification if someone types BaggageContext into a comment or PR?

ktoso:
I watch everything 😉

PR #403, opened 2020-11-12
Add env variable SOTO_CROSS_COMPILE to enable cross compile of package

adam-fowler:
This is a complete hack, but without https://bugs.swift.org/browse/SR-13761 being fixed soto-core requires the if statement at the base of the Package.swift. This causes cross compilation to fail as the Package.swift is read by macOS and the switch on which OS is being run fails.


adam-fowler:
Closing this as this is resolved on the v7.0 branch

PR #404, opened 2020-11-13
Allow loading credentials from profile and default

eneko:
AWS credentials file allows for using values from the "default" profile when these have not been overridden by another profile. This is, a profile that does not have a property defined, will use the default value for it.

This update fixes the code that loads credentials when a given profile is specified.

Example credentials file:
```
[default]
aws_session_token = TESTDEFAULT-SESSIONTOKEN
aws_access_key_id = TESTDEFAULT-AWSACCESSKEYID
aws_secret_access_key = TESTDEFAULT-AWSSECRETACCESSKEY

[test-profile]
aws_access_key_id = TESTPROFILE-AWSACCESSKEYID
aws_secret_access_key = TESTPROFILE-AWSSECRETACCESSKEY
```

Whit this configuration file, when `test-profile` profile is used, the value for `aws_session_token` should be `TESTDEFAULT-SESSIONTOKEN`.

More examples have been added as unit tests, with full code coverage.

### DIFF
--- a/Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
+++ b/Sources/SotoCore/Credential/ConfigFileCredentialProvider.swift
@@ -99,15 +99,18 @@ struct AWSConfigFileCredentialProvider: CredentialProvider {
             throw ConfigFileError.missingProfile(profile)
         }
 
-        guard let accessKeyId = config["aws_access_key_id"] else {
+        // Profile credentials can "borrow" values from the 'default' profile when they have not been overriden
+        let defaultConfig = parser.sections["default"]
+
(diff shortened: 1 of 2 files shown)


adam-fowler:
I'll look at this sometime this weekend. One question though, is this the same functionality as aws cli?

eneko:
@adam-fowler correct, that is how AWS CLI and AWS SAM work.

eneko:
Here is more documentation on the [AssumeRole](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) operation, and [Assume role provider in Boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#assume-role-provider).

It seems like an API call to AWS STS service is needed to retrieve temporary credentials. I'm going to go ahead and close this PR, and open an issue instead.

PR #407, opened 2020-11-17
[CI] Use install-swift-tool instead of mint

Cyberbeni:
install-swift-tool also builds tools from source, like mint but it includes caching and avoids the overhead of `brew install mint`

I believe that the current trigger means that it will generate a cache for every pull request. If we also add a trigger to run on pushes to the default branch then the pull requests could also load it from cache on the first run of every pull request. Source: https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -13,9 +13,10 @@ jobs:
       uses: actions/checkout@v1
       with:
         fetch-depth: 1
-    - name: Install Dependencies
-      run: |
-        brew install mint
-        mint install nicklockwood/swiftformat@0.46.3
+    - name: Install SwiftFormat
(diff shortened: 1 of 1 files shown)


Cyberbeni:
Looks like [resolving dependencies took an unusually long time](https://github.com/soto-project/soto-core/runs/1414375068?check_suite_focus=true) but that also [happened with mint before](https://github.com/soto-project/soto-core/runs/1332189380?check_suite_focus=true). But now if we rerun the sanity check, it should complete in a couple seconds since it would load from cache.

Cyberbeni:
This would also work on ubuntu, assuming that scripts/sanity.sh works on ubuntu

PR #415, opened 2020-12-11
Support credential_source in config file

adam-fowler:
Removed enum case `credentialSource` instead use `assumeRole` for both source_profile and credential_source. Instead of it holding a `StaticCredential` it now holds a `CredentialProviderFactory`. In the situation we are dealing with source_profile this is set to `.static()`. For `credential_source` we use one of `.environment`, `.ec2`, `.ecs`.

Add test, testing with environment credential


eneko:
Looks pretty good, @adam-fowler. I don't have an environment set up to test this feature, but looks like it should work well.

adam-fowler:
`testCredentialProviderSourceEnvironment` shows it getting source credentials from the environment. I could setup a full path, with fake EC2 credentials and a fake STS response but we have shown all the parts work so not sure that is necessary.

eneko (approved):

Issue #429, opened 2021-03-30
generateOutputShape throws when x-amz-executed-version is not $LATEST


saltzmanjoelh:
**Describe the bug**
Lambdas can be versioned. These versions are ints that are automatically incremented by 1 in the AWS Console. When executing a version that is not $LATEST generateOutputShape will throw an error about a type mismatch. 

**To Reproduce**
1. Create a Lambda function (called mytest)
2. Publish a version (1)
3. Update the code
4. Publish another version (2)
5. Create an alias (called myalias) and point it to version 1
6. Invoke the Lambda `Lambda.FunctionConfiguration.init(functionName: "mytest:myalias")

**Expected behavior**
An InvocationResponse should be received.

**Setup (please complete the following information):**
 - OS: macOS
 - Version of aws-sdk-swift: main branch 55c94c4
 - Authentication mechanism: local credentials file

**Additional context**
Given the key and value from the headers:
    - key : "x-amz-executed-version"
    - value : "1"
Around line 172 of AWSResponse.swift.
The value ("1") is converted to stringValue. Later down the code, we try to get the Double(stringValue), which works since it's "1". Finally, when `try decoder.unbox(dictionary, as: type)` is called, an error is thrown.

`throw DecodingError._typeMismatch(at: self.codingPath, expectation: type, reality: value)`
codingPath == CodingKeys(stringValue: "X-Amz-Executed-Version", intValue: nil)
type == String
value == 1 (not a string)

```swift 
        // add header values to output dictionary, so they can be decoded into the response object
        for (key, value) in self.headers {
            let headerParams = Output.headerParams
            if let index = headerParams.firstIndex(where: { $0.key.lowercased() == key.lowercased() }) {
                // check we can convert to a String. If not just put value straight into output dictionary
                guard let stringValue = value as? String else {
                    outputDict[headerParams[index].key] = value
                    continue
                }
                if let number = Double(stringValue) { // Problem is here when stringValue == "1"
                    outputDict[headerParams[index].key] = number.truncatingRemainder(dividingBy: 1) == 0 ? Int(number) : number
                } else if let boolean = Bool(stringValue) {
                    outputDict[headerParams[index].key] = boolean
                } else {
                    outputDict[headerParams[index].key] = stringValue
                }
            }
        }
```

*** Example Test ***

```swift
func testHeaderResponseDecodingStringWithIntegerValue() {
        struct Output: AWSDecodableShape {
[...]

PR #430, opened 2021-03-30
Add StringFlexibleDecodable for decoding headers values

adam-fowler:
When decoding JSON we first convert to a dictionary so we can add in additional values read from the headers and then push the resultant dictionary through the dictionary decoder. We have to guess whether the value is going to be decoded as a number, boolean or string as the decoder will throw a typeMismatch error. This is not ideal. So I have added a new type which encapsulates values extracted from headers and have that be convertible to `Bool`, `NSNumber` or `String`.

Add StringFlexibleDecodable for decoding headers values added to dictionaries before decoding with the DictionaryDecoder


adam-fowler:
This should resolve #429

adam-fowler:
If someone can think of a better name than `StringFlexibleDecodable` please say so

PR #431, opened 2021-04-12
Async/await

adam-fowler:
Related Soto PR is [here](https://github.com/soto-project/soto/pull/488)

Async/Await additions
- Added async versions `AWSClient.execute`
- Added `PaginatorSequence`. Unfortunately `AsyncSequence` functions like `map`, `reduce` etc are not available. Added my own versions of `reduce`
- Added async versions of `signURL` and `signHeaders`
- Temporarily added `EventLoopFuture.get` and `EventLoopPromise.completeWithAsync` until they appear in a NIO release
- Added `AsyncCredentialProvider`
- Added Async tests

Last tested with 2021-09-15 snapshot of Swift 5.5 compiler


Sources/SotoCore/AWSClient+Paginate+async.swift
```diff
@@ -0,0 +1,122 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Soto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=5.5) && $AsyncAwait
```
eneko, line 15: This is nice. How/where is `$AsyncAwait` defined?

eneko (commented):
This is looking really good, @adam-fowler , nice work.

(I haven't tried the code myself, only reviewed it here)

Is this ready to be merged in `main` or are we waiting for more changes?

adam-fowler:
@eneko I think this is ready to merge. Although I would prefer that AsyncSequence was more solid before merging. Currently all AsyncSequence functions are rethrows and are waiting on a swift evolution proposal for marking functions that call throwing protocol requirements rethrows instead of throws. 

Also, I'm not sure about having code dependent on unreleased swift features in the main branch. Might have a chat with swift guys about this.

Sources/SotoCore/AWSClient+Paginate+async.swift
adam-fowler: I just copied this from swift nio. Not sure where $AsyncAwait comes from exactly.

eneko:
Sounds reasonable to keep the PR open until the updates are released by the Swift team

ktoso:
Yeah I recommend holding off with the async sequence things _yet_, it might wobble around a bit in it's exact shape still.

Package.swift
```diff
@@ -25,10 +25,10 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
         .package(url: "https://github.com/apple/swift-metrics.git", "1.0.0"..<"3.0.0"),
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.16.1"),
-        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.7.2"),
-        .package(url: "https://github.com/apple/swift-nio-transport-services.git", from: "1.0.0"),
-        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.3.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", .upToNextMajor(from: "2.32.0")),
```
0xTim, line 28: Any particular reason to this being `upToNextMajor`? I seem to remember it moving to `from` in a previous PR. Either way the whole manifest should be consistent

0xTim (approved):
LGTM! Once AHC gets its async APIs merged in we should probably migrate the async APIs to use that to ensure tracing and logging work

Package.swift
adam-fowler: They were originally all `upToNextMajor`. I converted to `from` as it seemed more common. There is a possibility I did this after the async-await branch was created and then each time I would rebase with main it would jump about. Will revert to `from`

adam-fowler:
> LGTM! Once AHC gets its async APIs merged in we should probably migrate the async APIs to use that to ensure tracing and logging work

Unfortunately that's a breaking change. The http client protocol which is public includes functions that return `EventLoopFutures`. Slightly frustrating as basically nobody cares about writing their own client. Thought maybe someone might write one with `URLSession` for iOS. 

I could make the protocol disappear in the Smithy major update and the do as you suggested.

PR #432, opened 2021-04-14
Add `AWSClient.Options.errorLogLevel`

adam-fowler:
Also make retry logging trace

This change is in response to https://github.com/swift-server/guides/blob/main/log-levels.md


eneko (approved):
Looks good

PR #438, opened 2021-05-14
Remove hack to fix issues with AHC backpressure

adam-fowler:
Before AsyncHTTPClient v1.3.0 if you implemented your own `ResponseDelegate` the `EventLoopFuture` that `HTTPClient.execute` returns could complete before your delegate had finished processing the body parts. I had to add a hack to the `AWSClient` streaming response delegate to ensure we returned after all processing had been done by adding an additional promise to the delegate and then waiting on that promise completing before I could guarantee the delegate had done its work.

Now AHC v1.3.0 is out I can remove this hack.


eneko (approved):
It's always good when we get to remove some code :)

PR #439, opened 2021-05-19
Run request init and signing on a NIOThreadPool

adam-fowler:
We are doing this to take the generation of requests off the EventLoop as it not a cheap operation.

Add optional `NIOThreadPool` to `AWSClient.Options`
If thread pool is non-nil then use it for running the creation and signing of Requests


adam-fowler:
Closing. While this releases pressure on EventLoop. result tend to indicates overall this degrades performance

PR #443, opened 2021-05-27
Add support for tracing

0xTim:
Draft PR to see what things have been added that allowed us to support tracing with Soto.

This will be ongoing as we add more service support and missing pieces


Sources/SotoCore/AWSClient.swift
```diff
@@ -455,13 +459,34 @@ extension AWSClient {
         execute: @escaping (AWSHTTPRequest, EventLoop, Logger) -> EventLoopFuture<AWSHTTPResponse>,
         processResponse: @escaping (AWSHTTPResponse) throws -> Output,
         config: AWSServiceConfig,
-        logger: Logger = AWSClient.loggingDisabled,
+        context: LoggingContext,
```
adam-fowler, line 462: You've removed the default argument here so someone has to supply a context, but I guess this is a moot point as eventually we will be getting this from the task local storage

Sources/SotoCore/AWSClient.swift
```diff
@@ -455,13 +459,34 @@ extension AWSClient {
...
+            span.attributes.merge(serviceDetails.serviceAttributes)
+            operationSpan = span
+        }
+        var context = context
+        context.baggage = operationSpan?.baggage ?? context.baggage
+        operationSpan?.attributes["cloud.region"] = config.region.rawValue
+        operationSpan?.attributes["cloud.provider"] = config.region.rawValue
+        let peerURLComponents = URLComponents(string: config.endpoint)
+        operationSpan?.attributes.net.peer.ip = peerURLComponents?.host
+        operationSpan?.attributes.net.peer.port = peerURLComponents?.port
```
adam-fowler, line 485: I would have put all this code in `recordRequest`

Sources/SotoCore/AWSClient.swift
```diff
@@ -455,13 +459,34 @@ extension AWSClient {
         execute: @escaping (AWSHTTPRequest, EventLoop, Logger) -> EventLoopFuture<AWSHTTPResponse>,
         processResponse: @escaping (AWSHTTPResponse) throws -> Output,
         config: AWSServiceConfig,
-        logger: Logger = AWSClient.loggingDisabled,
+        context: LoggingContext,
         on eventLoop: EventLoop? = nil
     ) -> EventLoopFuture<Output> {
+        var operationSpan: Span?
+        if let serviceDetails = context.baggage[SotoServiceBaggageKey.self] {
```
adam-fowler, line 466: Are the service details added by the service calling `AWSClient.execute`

Sources/SotoCore/Tracing/Baggage+ServiceName.swift
```diff
@@ -0,0 +1,19 @@
+import Baggage
+import Tracing
+
+public enum SotoServiceBaggageKey: Baggage.Key {
+    public typealias Value = SotoServiceBaggage
+}
+
+public struct SotoServiceBaggage {
+    public let serviceName: String
+    public let databaseSystem: String?
```
adam-fowler, line 10: `databaseSystem` seems very specific to certain services. Should this not go into `serviceAttributes`

adam-fowler (commented):
I know a lot of this isn't a finalised PR but I've added a couple comments/questions

Sources/SotoCore/AWSClient.swift
slashmo: Yes, I purposefully broke the API here because I don't intend/recommend someone else to use it. With `TaskLocal`s landing soon we'll likely remove `LoggingContext` entirely, as the feature of automatically logging metadata derived from `Baggage` could live in `Logging` itself.

slashmo:
@adam-fowler Setting span attributes is going to be tricky in Soto. 

As you can see in the OpenTelemetry spec there are tons of attributes that are recommended for an AWS SDK not only dependent on the service but also on the exact operation: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md. `DynamoDB.BatchGetItem` e.g. are expected to add an attribute defining all table names (`aws.dynamodb.table_names`). Granted, these highly specific attributes are not required, though might provide useful insights.

Passing the attributes through `Baggage` was definitely a last-minute hack and in my opinion, not an ideal solution, as it required us to still make changes to the generated code in the Soto fork: https://github.com/brokenhandsio/soto/commit/6ce0b658944785649c5cced805cab7f0c3bc5c7e

Do you have an idea on how to make this happen while still auto-generating the services? What I thought about roughly would be to have the generated code delegate retrieving span attributes to some object that could be extended for specific use-cases, like DynamoDB batch requests. These extensions would then live alongside the generated code much like the `Codable` extensions for DynamoDB.

Sources/SotoCore/Tracing/Baggage+ServiceName.swift
slashmo: Please see my comment below: https://github.com/soto-project/soto-core/pull/443#issuecomment-850554367

Package.swift
```diff
@@ -25,10 +25,12 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
         .package(url: "https://github.com/apple/swift-metrics.git", "1.0.0"..<"3.0.0"),
+        .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "0.1.2"),
         .package(url: "https://github.com/apple/swift-nio.git", .upToNextMajor(from: "2.16.1")),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", .upToNextMajor(from: "2.7.2")),
         .package(url: "https://github.com/apple/swift-nio-transport-services.git", .upToNextMajor(from: "1.0.0")),
-        .package(url: "https://github.com/swift-server/async-http-client.git", .upToNextMajor(from: "1.3.0")),
+//        .package(url: "https://github.com/swift-server/async-http-client.git", .upToNextMajor(from: "1.3.0")),
+        .package(url: "https://github.com/slashmo/async-http-client.git", .branch("feature/tracing")),
```
pokryfka, line 33: would be great to have tracing in AHC, a lot of libraries depend on that

slashmo: Yep, that’s only a matter of time. Though for libraries it doesn’t make sense to merge in support yet because they would need to accept a `LoggingContext` in their APIs until Baggage will be passed through a task-local (Swift 5.5).

adam-fowler:
> @adam-fowler Setting span attributes is going to be tricky in Soto.
> 
I had a look through the other SDKs to see what they were doing and basically none of the SDKs seem to be have any support for tracing. What they have instead is a separate module that wraps the SDKs and is used to setup spans etc . I'm not sure exactly how it works. I'm going to investigate a little further.

adam-fowler:
closing. Solution is available on 7.x.x branch

PR #446, opened 2021-07-09
Add support for waiters using JMESPath

adam-fowler:
Waiters are a client-side abstraction used to poll a resource until a desired state is reached, or until it is determined that the resource will never enter into the desired state. This is a common task when working with services that are eventually consistent like Amazon S3 or services that asynchronously create resources like Amazon EC2. Writing logic to continuously poll the status of a resource can be cumbersome and error-prone. The goal of waiters is to move this responsibility out of customer code and onto Soto.

- Added `AWSClient.waitUntil` function which polls a resource until a matcher returns a positive
- Implemented various matchers. Include path (matching value using JMESPath), error status, error code, anyPath and allPath for multiple values returned by a path

See the Smithy spec for a more in-depth description of Waiters https://awslabs.github.io/smithy/1.0/spec/waiters.html

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -44,7 +44,7 @@ let package = Package(
             .product(name: "NIOSSL", package: "swift-nio-ssl"),
             .product(name: "NIOTransportServices", package: "swift-nio-transport-services"),
             .product(name: "NIOFoundationCompat", package: "swift-nio"),
-            .product(name: "JMESPath", package: "jmespath.swift")
+            .product(name: "JMESPath", package: "jmespath.swift"),
         ]),
         .target(name: "SotoCrypto", dependencies: []),
         .target(name: "SotoSignerV4", dependencies: [
--- a/Package.swift
+++ b/Package.swift
@@ -29,7 +29,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-ssl.git", .upToNextMajor(from: "2.7.2")),
         .package(url: "https://github.com/apple/swift-nio-transport-services.git", .upToNextMajor(from: "1.0.0")),
         .package(url: "https://github.com/swift-server/async-http-client.git", .upToNextMajor(from: "1.3.0")),
-        .package(url: "https://github.com/adam-fowler/jmespath.swift.git", .upToNextMajor(from: "0.2.0")),
+        .package(url: "https://github.com/adam-fowler/jmespath.swift.git", .branch("jmes-property-wrapper")),
     ],
     targets: [
         .target(name: "SotoCore", dependencies: [
--- a/Package.swift
+++ b/Package.swift
@@ -29,6 +29,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-ssl.git", .upToNextMajor(from: "2.7.2")),
         .package(url: "https://github.com/apple/swift-nio-transport-services.git", .upToNextMajor(from: "1.0.0")),
         .package(url: "https://github.com/swift-server/async-http-client.git", .upToNextMajor(from: "1.3.0")),
+        .package(url: "https://github.com/adam-fowler/jmespath.swift.git", .upToNextMajor(from: "0.2.0")),
     ],
     targets: [
         .target(name: "SotoCore", dependencies: [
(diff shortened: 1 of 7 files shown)


eneko (approved):
👏

Package.swift
```diff
@@ -25,10 +25,11 @@ let package = Package(
...
         .package(url: "https://github.com/apple/swift-metrics.git", "1.0.0"..<"3.0.0"),
-        .package(url: "https://github.com/apple/swift-nio.git", .upToNextMajor(from: "2.16.1")),
-        .package(url: "https://github.com/apple/swift-nio-ssl.git", .upToNextMajor(from: "2.7.2")),
-        .package(url: "https://github.com/apple/swift-nio-transport-services.git", .upToNextMajor(from: "1.0.0")),
-        .package(url: "https://github.com/swift-server/async-http-client.git", .upToNextMajor(from: "1.3.0")),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.16.1"),
+        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.7.2"),
+        .package(url: "https://github.com/apple/swift-nio-transport-services.git", from: "1.0.0"),
+        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.3.0"),
+        .package(url: "https://github.com/adam-fowler/jmespath.swift.git", from: "1.0.0"),
```
adam-fowler, line 32: This PR brings in a new dependency

Sources/SotoCore/Waiters/Matcher.swift
```diff
@@ -0,0 +1,165 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import JMESPath
+import NIO
+import NIOHTTP1
+
+/// Protocol for matchers used in waiters.
+///
+/// A matcher returns whether returned value from API call matches a certain state
```
0xTim, line 21: Nit - grammar
```suggestion
/// A matcher returns whether the returned value from ah API call matches a certain state
```

Sources/SotoCore/Waiters/Matcher.swift
```diff
@@ -0,0 +1,165 @@
...
+        switch result {
+        case .success:
+            return true
+        case .failure:
+            return false
+        }
+    }
+}
+
+/// Match whether a call return a specific HTTP response code
```
0xTim, line 123: Nit - grammar
```suggestion
/// Match whether a call returns a specific HTTP response code
```

Sources/SotoCore/Waiters/Matcher.swift
```diff
@@ -0,0 +1,165 @@
...
+            }
+        case .failure(let error as AWSRawError):
+            return error.context.responseCode.code == self.expectedStatus
+        case .failure:
+            return false
+        }
+    }
+}
+
+/// Match whether a call returned a specific error code
```
0xTim, line 149: Nit -  align tense

```suggestion
/// Match whether a call returns a specific error code
```

Tests/SotoCoreTests/WaiterTests.swift
```diff
@@ -0,0 +1,325 @@
...
+//===----------------------------------------------------------------------===//
+
+import AsyncHTTPClient
+import NIO
+import SotoCore
+import SotoTestUtils
+import XCTest
+
+class WaiterTests: XCTestCase {
+    static var awsServer: AWSTestServer!
```
0xTim, line 22: Out of interest why are these static properties instead of instance properties? Won't that stop you being able to run the tests in parallel?

0xTim (approved):
In general LGTM. A few nits around comments and a couple of code queries

Sources/SotoCore/Waiters/AWSClient+Waiter.swift
```diff
@@ -0,0 +1,156 @@
...
+
+import Dispatch
+import Foundation
+import NIO
+
+// MARK: Waiters
+
+extension AWSClient {
+    /// Waiter state
+    public enum WaiterState {
```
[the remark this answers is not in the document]
adam-fowler: Waiters can only ever be set to successful, failed, or to retry. You can see this in the Smithy waiter spec https://awslabs.github.io/smithy/1.0/spec/waiters.html

Tests/SotoCoreTests/WaiterTests.swift
adam-fowler: Yeah I guess so. I've never really run the tests in parallel. Will change

PR #448, opened 2021-07-19
Add host prefix to AWSRequest

adam-fowler:
A number of services have API calls that add a prefix to the host name (chime, location, s3control etc). This PR implements this by adding an optional `hostPrefix` parameter to all the `AWSClient.execute` functions. If set it prepends this value to the hostname. It is also possible to have input shape members affect this prefix. See `S3Control` that prepends the hostname with the account number of the user.

See related Soto PR https://github.com/soto-project/soto/pull/507

### DIFF
--- a/Sources/SotoCore/Message/AWSRequest.swift
+++ b/Sources/SotoCore/Message/AWSRequest.swift
@@ -120,11 +120,12 @@ extension AWSRequest {
         path: String,
         httpMethod: HTTPMethod,
         input: Input,
+        hostPrefix: String? = nil,
         configuration: AWSServiceConfig
     ) throws {
         var headers = HTTPHeaders()
         var path = path
-        var hostPrefix = Input._hostPrefix
+        var hostPrefix = hostPrefix
         var body: Body = .empty
         var queryParams: [(key: String, value: Any)] = []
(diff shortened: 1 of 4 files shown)


Sources/SotoCore/Message/AWSRequest.swift
```diff
@@ -169,7 +170,8 @@ extension AWSRequest {
                     path = path
                         .replacingOccurrences(of: "{\(location)}", with: Self.urlEncodePathComponent(String(describing: value)))
                         .replacingOccurrences(of: "{\(location)+}", with: Self.urlEncodePath(String(describing: value)))
-
+                    hostPrefix = hostPrefix?
+                        .replacingOccurrences(of: "{\(location)}", with: Self.urlEncodePathComponent(String(describing: value)))
```
adam-fowler, line 174: I would prefer to add a new location `.host` which affects the hostname but that would be a breaking change. So I'm using `.uri` which kinda makes sense

0xTim: Yeah I think it's about as good as it can be

0xTim (approved):

PR #452, opened 2021-08-07
Stream writer

adam-fowler:
This introduces a new way to stream data into a request. Instead of asking the user for another buffer when needed via a callback a stream writer allows you to push data to a request when you want to. 
```swift
let streamWriter = AWSStreamWriter(length: 16384, on: eventLoop)
let futureResult = s3.putObject(body: .streamWriter(streamWriter), bucket: "mybucket", key: "mykey") 
streamWriter.write(.byteBuffer(buffer1))
streamWriter.write(.byteBuffer(buffer2))
streamWriter.write(.end)
try futureResult.wait()
```

In the example above we ignore the return value of `AWSStreamWriter.write`but it does return an `EventLoopFuture` which can be used to apply back pressure.

`AWSStreamWriters` can be chained together so the results of one can be processed by the next. This is used internally to implement the S3 signed streaming. The `AWSStreamWriter` results are sent to a `S3StreamWriter` which then sends its results onto the AsyncHTTPClient StreamWriter.

Now we have two full streaming solutions implemented. I hope to rewrite `StreamReader` using the new `AWSStreamWriter` instead of having it talk directly to AsyncHTTPClient. With this we can remove `S3ChunkedStreamReader` and use the chaining of StreamWriters for S3 signed `StreamReader` streams.

### DIFF
--- a/Sources/SotoCore/HTTP/S3StreamWriter.swift
+++ b/Sources/SotoCore/HTTP/S3StreamWriter.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Soto for AWS open source project
 //
-// Copyright (c) 2017-2020 the Soto project authors
+// Copyright (c) 2017-2021 the Soto project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -63,12 +63,12 @@ class S3StreamWriter: StreamWriterProtocol {
         self.tailBuffer.writeString("\r\n")
     }
 
-    /// Update HTTP headers. Add "Content-encoding" header. The "x-amz-decoded-content-length" header is added earlier in AWSRequest when
+    /// Update HTTP headers. Add "content-encoding" header. The "x-amz-decoded-content-length" header is added earlier in AWSRequest when
     /// initiating the signing process
     /// - Parameter headers: headers to update
     func updateHeaders(headers: HTTPHeaders) -> HTTPHeaders {
         var headers = headers
-        headers.add(name: "Content-Encoding", value: "aws-chunked")
+        headers.add(name: "content-encoding", value: "aws-chunked")
         return headers
     }
 
--- a/Sources/SotoCore/HTTP/S3StreamWriter.swift
+++ b/Sources/SotoCore/HTTP/S3StreamWriter.swift
@@ -52,7 +52,7 @@ class S3StreamWriter: StreamWriterProtocol {
         self.eventLoop = eventLoop
         self.writerPromise = eventLoop.makePromise()
         self.finishedPromise = eventLoop.makePromise()
-        
+
         self.signer = signer
         self.signingData = seedSigningData
         // have separate buffers so we aren't allocating 128k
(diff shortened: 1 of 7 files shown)


Sources/SotoCore/HTTP/S3StreamWriter.swift
```diff
@@ -0,0 +1,140 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Soto for AWS open source project
+//
+// Copyright (c) 2017-2020 the Soto project authors
```
0xTim, line 5: This probably needs to be updated
```suggestion
// Copyright (c) 2017-2021 the Soto project authors
```

Sources/SotoCore/HTTP/S3StreamWriter.swift
```diff
@@ -0,0 +1,140 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import SotoSignerV4
+
+class S3StreamWriter: StreamWriterProtocol {
+    /// Working buffer size
+    static let bufferSize: Int = 64 * 1024
```
0xTim, line 20: Out of curiosity why have we picked these numbers? If it's an S3 thing we should probably link to the docs

Sources/SotoCore/HTTP/S3StreamWriter.swift
```diff
@@ -0,0 +1,140 @@
...
+        self.tailBuffer = byteBufferAllocator.buffer(capacity: Self.endOfLineLength)
+        self.tailBuffer.writeString("\r\n")
+    }
+
+    /// Update HTTP headers. Add "Content-encoding" header. The "x-amz-decoded-content-length" header is added earlier in AWSRequest when
+    /// initiating the signing process
+    /// - Parameter headers: headers to update
+    func updateHeaders(headers: HTTPHeaders) -> HTTPHeaders {
+        var headers = headers
+        headers.add(name: "Content-Encoding", value: "aws-chunked")
```
0xTim, line 71: nit - headers _should_ be lowercased (at least in HTTP/1 - it's a must in HTTP/2 IIRC) and are lowercased internally so I suggest we follow that. Thoughts?

```suggestion
        headers.add(name: "content-Encoding", value: "aws-chunked")
```

Sources/SotoCore/HTTP/S3StreamWriter.swift
```diff
@@ -0,0 +1,140 @@
...
+        self.signingData = seedSigningData
+        // have separate buffers so we aren't allocating 128k
+        self.byteBufferAllocator = byteBufferAllocator
+        self.headerBuffer = byteBufferAllocator.buffer(capacity: Self.maxHeaderSize)
+        self.workingBuffer = byteBufferAllocator.buffer(capacity: Self.bufferSize)
+        self.tailBuffer = byteBufferAllocator.buffer(capacity: Self.endOfLineLength)
+        self.tailBuffer.writeString("\r\n")
+    }
+
+    /// Update HTTP headers. Add "Content-encoding" header. The "x-amz-decoded-content-length" header is added earlier in AWSRequest when
```
0xTim, line 66: See below

```suggestion
    /// Update HTTP headers. Add "content-encoding" header. The "x-amz-decoded-content-length" header is added earlier in AWSRequest when
```

Sources/SotoCore/HTTP/StreamWriter.swift
```diff
@@ -0,0 +1,126 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Soto for AWS open source project
+//
+// Copyright (c) 2017-2020 the Soto project authors
```
0xTim, line 5: Nit

```suggestion
// Copyright (c) 2017-2021 the Soto project authors
```

Sources/SotoCore/HTTP/StreamWriter.swift
```diff
@@ -0,0 +1,126 @@
...
+    /// Write to stream writer
+    /// - Parameter result: ByteBuffer or `.end`
+    /// - Returns: Returns when EventLoopFuture for when value has been written
+    @discardableResult public func write(_ result: StreamWriterResult) -> EventLoopFuture<Void> {
+        let futureResult = self.write(result, on: self.eventLoop)
+        futureResult.cascadeFailure(to: self.finishedPromise)
+        return futureResult
+    }
+
+    /// Update headers. Add "Transfer-encoding" header if we don't have a steam size
```
0xTim, line 106: As above
```suggestion
    /// Update headers. Add "transfer-encoding" header if we don't have a steam size
```

Sources/SotoCore/HTTP/StreamWriter.swift
```diff
@@ -0,0 +1,126 @@
...
+    /// - Parameter headers: headers to update
+    func updateHeaders(headers: HTTPHeaders) -> HTTPHeaders {
+        var headers = headers
+        // add "Transfer-Encoding" header if streaming with unknown size
+        if self.length == nil {
+            headers.add(name: "Transfer-Encoding", value: "chunked")
```
0xTim, line 112: As above 
```suggestion
            headers.add(name: "transfer-encoding", value: "chunked")
```

0xTim (approved):
LGTM a few queries/minor nits

Sources/SotoCore/HTTP/S3StreamWriter.swift
adam-fowler: This is the recommended size from AWS. See https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html

Sources/SotoCore/HTTP/S3StreamWriter.swift
adam-fowler: If we're gonna lowercase the "c" we should probably also do the same to the "e" 😝

Sources/SotoCore/HTTP/S3StreamWriter.swift
adam-fowler: done

Sources/SotoCore/HTTP/S3StreamWriter.swift
0xTim: I think we should add that as a comment

PR #454, opened 2021-08-12
Add `omitSecurityToken` parameter to `AWSSigner.signHeaders()`

adam-fowler:
If `omitSecurityToken` is set to true the `x-amz-security-token` header will not be included in the V4 signature. This is required for IoT MQTT connections. The interface replicates the same interface used in the new AWS C runtime.

### DIFF
--- a/Sources/SotoSignerV4/signer.swift
+++ b/Sources/SotoSignerV4/signer.swift
@@ -73,18 +73,26 @@ public struct AWSSigner {
     }
 
     /// Generate signed headers, for a HTTP request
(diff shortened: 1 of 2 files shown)


moritzsternemann:
Thanks for the quick fix @adam-fowler 🚀 

Though I think for use with AWS IoT we need the same functionality in the `AWSSigner.signURL()` method so the result can be passed as the [`webSocketURLPath` parameter of `MQTTClient.Configuration`s initializer](https://github.com/adam-fowler/mqtt-nio/blob/main/Sources/MQTTNIO/MQTTConfiguration.swift#L44).

adam-fowler:
@moritzsternemann sorry completely missed that. Will add it in today

adam-fowler:
Hey @moritzsternemann do you want to try again.

moritzsternemann:
Thanks again @adam-fowler! The connection works as expected when setting `omitSecurityToken` to true.

PR #457, opened 2021-08-25
Add support for Endpoint discovery

adam-fowler:
DynamoDB and TimeStream both use endpoint discovery to provide endpoints to the user. For DynamoDB this is an optional feature but for TimeStream this is a requirement. This PR adds endpoint discovery internal to Soto instead of expecting the user to do this.

The code adds in five new versions of `AWSClient.execute` with additional parameter `endpointDiscovery`. This type contains storage for a cached endpoint and a closure to call to get a new endpoint. 

There are two types of endpoint acquisition: one where you are required to use the endpoint returned by endpoint discovery (TimeStream) and one where you can use the standard endpoint while endpoint discovery is in progress (DynamoDB). Both of these are supported. Given the DynamoDB calls are not dependent on the endpoint discovery to finish, the impact on DynamoDB calls is minimal. 

Because DynamoDB doesn't require endpoint discovery it is disabled by default. If you would like to enable it there is a new `AWSServiceConfig.Option` `enableEndpointDiscovery`.

Related Soto PR is https://github.com/soto-project/soto/pull/517

### DIFF
--- a/Sources/SotoCore/AWSClient+EndpointDiscovery.swift
+++ b/Sources/SotoCore/AWSClient+EndpointDiscovery.swift
@@ -47,7 +47,7 @@ extension AWSClient {
                     on: eventLoop
                 )
             },
-            isDisabled: serviceConfig.options.contains(.disableEndpointDiscovery),
+            isEnabled: serviceConfig.options.contains(.enableEndpointDiscovery),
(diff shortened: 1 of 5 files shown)


0xTim (approved):
LGTM

PR #459, opened 2021-09-16
Replace import NIO with import NIOCore

adam-fowler:
Needed to add a couple of import NIOPosix calls as well, hopefully can remove those at some point


0xTim (approved):
LGTM

PR #461, opened 2021-09-17
Add AWSPayload.asyncSequence

adam-fowler:
This adds AWSPayload.asyncSequence which allows you to provide an AsyncSequence of ByteBuffers to a request.

This will merge into the bigger async-await PR #431

### DIFF
--- a/Sources/SotoCore/AWSShapes/Payload+async.swift
+++ b/Sources/SotoCore/AWSShapes/Payload+async.swift
@@ -1,0 +1,48 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Soto for AWS open source project
+//
+// Copyright (c) 2017-2021 the Soto project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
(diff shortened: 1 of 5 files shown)


Sources/SotoCore/AWSShapes/Payload+async.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=5.5)
+
+import NIOCore
+
+@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+extension AWSPayload {
+    /// Construct a stream payload from a `NIOFileHandle`
```
0xTim, line 21: Only comment after a quick read through of this is that it's not clear where `NIOFileHandle` comes from or is used in this extension

0xTim (approved):
One comment but LGTM

Sources/SotoCore/AWSShapes/Payload+async.swift
adam-fowler: Comment that came from cut and paste. I should have removed. 👍

PR #462, opened 2021-09-19
Fix compile issue with swift-nio in Xcode 13 GM

luispadron:
Because of an issue with Xcode 13 GM SDK, swift-nio requires a workaround to compile (https://github.com/apple/swift-nio/issues/1955). This has been released in `2.32.3`.


adam-fowler:
This change might take a few days to appear in a release. I am in the process of merging in the async await code. 

In the meantime you can just update your dependencies and it'll get the latest version of swift-nio. 

Outside of this issue I believe there are other swift concurrency compile issues with Xcode GM. I ended up using the latest version of swift from swift.org to get things working

PR #463, opened 2021-09-22
Wrap concurrency code in canImport(_Concurrency)

adam-fowler:
See https://github.com/apple/swift-nio/pull/1966#discussion_r714110863

### DIFF
--- a/Tests/SotoCoreTests/TimeStampTests.swift
+++ b/Tests/SotoCoreTests/TimeStampTests.swift
@@ -53,7 +53,7 @@ class TimeStampTests: XCTestCase {
         let a = A(date: Date(timeIntervalSince1970: 23_984_978_378))
         var request: AWSRequest?
         XCTAssertNoThrow(request = try AWSRequest(operation: "test", path: "/", httpMethod: .GET, input: a, configuration: createServiceConfig()))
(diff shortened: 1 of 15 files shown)


Tests/SotoCoreTests/TimeStampTests.swift
```diff
@@ -53,7 +53,7 @@ class TimeStampTests: XCTestCase {
         let a = A(date: Date(timeIntervalSince1970: 23_984_978_378))
         var request: AWSRequest?
         XCTAssertNoThrow(request = try AWSRequest(operation: "test", path: "/", httpMethod: .GET, input: a, configuration: createServiceConfig()))
-        #if os(Linux) && compiler(>=5.5)
+        #if os(Linux) && compiler(>=5.5) && canImport(_Concurrency)
```
0xTim, line 56: Does this one need to change? It doesn't look like the code in the block actually relies on concurrency and this is more a 5.5 quirk

0xTim (approved):
On e query otherwise LGTM

Tests/SotoCoreTests/TimeStampTests.swift
adam-fowler: good catch

PR #465, opened 2021-10-01
Don't throw errors on failure to decode additional fields of AWS error

adam-fowler:
When decoding an error returned by AWS, we attempt to extract any fields from the error that are non-standard and place them in a `Dictionary<String, String> called `additionalFields`. This ends up being not such a good idea because it assumes all additional fields will be strings. This change means if we decode an error which has a non string type additional field it will not throw an error and still create the error type expected.

This should fix https://github.com/soto-project/soto/issues/532


0xTim (approved):
LGTM

PR #478, opened 2021-11-11
Make AWSHTTPClient internal to project

adam-fowler:
So we can change it without requiring a major release
This will mean AsyncHTTPClient is the only HTTP client you can use with Soto


adam-fowler:
Cannot merge this until 6.x.x branch is merged

PR #479, opened 2021-11-11
Remove retry on NIOConnectionError as AsyncHTTPClient does this for us

adam-fowler:
Changed testCustomRetryPolicy test as it depended on retrying on NIOConnectionError

### DIFF
--- a/Sources/SotoCore/RetryPolicy.swift
+++ b/Sources/SotoCore/RetryPolicy.swift
@@ -93,10 +93,10 @@ extension StandardRetryPolicy {
                 }
             }
             return .dontRetry
-#if DEBUG
+        #if DEBUG
         case let httpClientError as HTTPClientError where httpClientError == .remoteConnectionClosed:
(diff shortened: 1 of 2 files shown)


0xTim (approved):

Sources/SotoCore/RetryPolicy.swift
```diff
@@ -93,10 +93,10 @@ extension StandardRetryPolicy {
                 }
             }
             return .dontRetry
+#if DEBUG
```
0xTim, line 96: Is there a reason this is wrapped in `#DEBUG`?

adam-fowler: Commonly when debugging AWS would close the connection because the client didn't respond quick enough. To avoid things failing I added the retry on connect. But you can't have this in a release situation because when you receive a connection close error you don't know if the request got through, you just didn't receive a response. If you retry you could end up running the operation twice. This is a problem if the operation is non-idempotent. @fabianfett was one to point this out to me.

PR #489, opened 2022-01-12
Revert availability check changes for earlier platform versions

adam-fowler:
Given the availability checks are causing issue in older Xcodes I'm thinking of reverting these changes

### DIFF
--- a/Sources/SotoCore/AWSShapes/Payload+async.swift
+++ b/Sources/SotoCore/AWSShapes/Payload+async.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if compiler(>=5.5) && canImport(_Concurrency)
+#if compiler(>=5.5.2) && canImport(_Concurrency)
 
 import NIOCore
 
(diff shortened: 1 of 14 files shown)


0xTim (approved):
LGTM - I think the other way around it is to duplicate everything and wrap it in the 5.5.2 check a la NIO, but that's a pain any why Vapor hasn't done it yet

adam-fowler:
> LGTM - I think the other way around it is to duplicate everything and wrap it in the 5.5.2 check a la NIO, but that's a pain any why Vapor hasn't done it yet

I would have to duplicate all the service functions as well so I think that is a no go.

I do have one advantage, I am in the process of doing a major version release. So my thoughts are revert this change for 5.x.x and require swift 5.5.2 (in effect requiring Xcode 13.2) for 6.x.x

0xTim:
Is there a way to require 5.5.2 short of wrapping everything in the compiler directive? IIRC package tools version only does minor versions

adam-fowler:
> Is there a way to require 5.5.2 short of wrapping everything in the compiler directive? IIRC package tools version only does minor versions

I'm already have a compiler directive for checking for swift 5.5. I was just going to change that to 5.5.2. 
Anyway I don't want people to not be able to use 6.x.x because they are using an older version of swift. I'm just making the concurrency code not available.

adam-fowler:
Actually this is the main branch (want to add the 5.5.2 check instead)

PR #494, opened 2022-03-15
Fix empty query protocol requests

adam-fowler:
Body should still contain Version and Action keys. It appears some services don't care but SES does see https://github.com/soto-project/soto/issues/580


0xTim (approved):
LGTM

PR #499, opened 2022-03-21
Add Http checksum calculation

adam-fowler:
Add support for calculating HTTP checksum. Checksums available include CRC32, CRC32C, SHA1 and SHA256


0xTim (approved):
So overall looks good.

I'm assuming you checked for a Swift implementation of CRC32 and CRC32c? I found https://github.com/my-mail-ru/swift-CRC32/blob/master/Sources/crc32.swift which looks like it uses the same lookup table, just called from Swift. Of course I don't know how well it looks and it might not cover everything we need (e.g. CRC32c might not be an easy port). However worth asking given the SSWG stance on C code

As an aside, I notice we're implementing our own Crypto off Linux. IIRC that was to avoid forcing people to 5.2, but we can include it now for the next major version right?

adam-fowler:
> So overall looks good.
> 
> I'm assuming you checked for a Swift implementation of CRC32 and CRC32c? I found https://github.com/my-mail-ru/swift-CRC32/blob/master/Sources/crc32.swift which looks like it uses the same lookup table, just called from Swift. Of course I don't know how well it looks and it might not cover everything we need (e.g. CRC32c might not be an easy port). However worth asking given the SSWG stance on C code
> 
> As an aside, I notice we're implementing our own Crypto off Linux. IIRC that was to avoid forcing people to 5.2, but we can include it now for the next major version right?

I used the C implementation from zlib because it is always going to be faster (no array bounds checking, on a very tight loop). zlib doesn't include a crc32c implementation though, even though it is almost the same. I could have implemented in swift, it is a simple operation, but felt speed was more important. It could be running on very large buffers.

Regarding the crypto. Linux uses swift-crypto and macOS uses CommonCrypto. I was going to get rid of the CommonCrypto implementation once we do a full async-await implementation.

adam-fowler:
Replaced the C CRC32 calculation with a Swift version (without performance hit!)

0xTim (approved):

PR #501, opened 2022-05-12
Replace usage of Foundation.Data with AWSBase64Data

adam-fowler:
In all cases where `Data` is used it is serialised as base64 data. So instead of storing a `Data` which is not thread safe, store the generated base64 data.


0xTim (approved):
LGTM

PR #505, opened 2022-05-29
Remove public symbols that shouldn't be public

adam-fowler:
- DictionaryDecoder
- Base64
- ConfigFile


Sources/SotoCore/Utils/base64.swift
```diff
@@ -70,8 +70,7 @@
 
 // MARK: - Extensions -
 
-public extension String {
-    @inlinable
+extension String {
```
0xTim, line 73: This is no longer `@inlinable` - is that deliberate?

0xTim (approved):
One query otherwise LGTM

Sources/SotoCore/Utils/base64.swift
adam-fowler: @inlinable is for public symbols. It basically another module might want to inline this symbol

PR #511, opened 2022-07-07
Fixup warning for incoming NIO release

adam-fowler:


0xTim (approved):
LGTM - annoying that Atomics doesn't support `Sendable`! You would have thought it would be first in the list!

adam-fowler:
There is a commit that makes UnsafeAtomic Sendable. For some reason they haven't made ManagedAtomic Sendable

Issue #523, opened 2022-10-11
AWSSigner processURL does not handle query parameters with no equal sign or value correctly


mildm8nnered:
**Describe the bug**

So I think there's an issue with the implementation of `processURL` in `AWSSigner`. Consider a request to begin a multipart upload, with bucket "MyBucket", the path according to Amazon should be "MyBucket?uploads".

But for signing, Amazon considers the encoded query string to be `uploads=`, and the `=` is important. Amazon will fail signatures that do not use that form as the second line of the canonical request string constructed during signing.

The issue is that `AWSSigner.processURL()` does not add the `=` in this case. At or around line 65, I think

```
        let urlQueryString = urlComponents.queryItems?
            .sorted {
                if $0.name < $1.name { return true }
                if $0.name > $1.name { return false }
                guard let value1 = $0.value, let value2 = $1.value else { return false }
                return value1 < value2
            }
            .map { item in item.value.map { "\(item.name)=\($0.uriEncode())" } ?? item.name }
            .joined(separator: "&")
```

really wants to be 

```
        let urlQueryString = urlComponents.queryItems?
            .sorted {
                if $0.name < $1.name { return true }
                if $0.name > $1.name { return false }
                guard let value1 = $0.value, let value2 = $1.value else { return false }
                return value1 < value2
            }
            .map { item in item.value.map { "\(item.name)=\($0.uriEncode())" } ?? "\(item.name)=" }
            .joined(separator: "&")
```

Assuming I'm right, it's really surprising to me that this hasn't come up before. I grabbed `soto-s3-file-transfer`, but looking there, the begin multipart request gets passed through another very similar piece of code, in AWSRequest.swift (around line 285), that performs an almost identical operation, but taking care of nil values correctly:

```
        // Set query params. Percent encode these ourselves as Foundation and AWS disagree on what should be percent encoded in the query values
        // Also the signer doesn't percent encode the queries so they need to be encoded here
        if queryParams.count > 0 {
            let urlQueryString = queryParams
                .map { (key: $0.key, value: "\($0.value)") }
                .sorted {
                    // sort by key. if key are equal then sort by value
                    if $0.key < $1.key { return true }
                    if $0.key > $1.key { return false }
                    return $0.value < $1.value
                }
                .map { "\($0.key)=\(Self.urlEncodeQueryParam($0.value))" }
                .joined(separator: "&")
            urlComponents.percentEncodedQuery = urlQueryString
        }
```

**To Reproduce**

[...]

PR #524, opened 2022-10-16
When a query parameter has no value, output is as `key=`

mildm8nnered:
When a URL query parameter has no value, AWS expects it to be presented as `key=`. processURL did not add the `=` sign.

See https://github.com/soto-project/soto-core/issues/523


adam-fowler:
From a quick look the change looks fine. I'm on holiday just now. Will review properly when I return in 10 days time.

adam-fowler (approved):

PR #529, opened 2022-11-11
Add .spi.yml to point to correct documentation

adam-fowler:
Add .spi.yml to get Swift Package Index to add documentation link


adam-fowler:
@finestructure is this correct?

finestructure:
The syntax is correct but it looks like you copied this with curly quotes instead of `"` – from our announcement post! I need to fix that!

0xTim (changes requested):
As @finestructure said

adam-fowler:
> As @finestructure said

fixed

0xTim (approved):
👍

PR #534, opened 2023-01-02
Add support for endpoint variants (FIPS, dualstack)

adam-fowler:
This adds support for setting additional endpoints for a service. It supports adding
- dual stack endpoints which support IPv6 and IPv4
- FIPS endpoints which support the US government FIPS requirements (ie at least TLS 1.2)

Adds two new options to `AWSServiceConfig.Options`: `.useFipsEndpoint`, `.useDualStackEndpoint`


Sources/SotoCore/Doc/Region.swift
```diff
@@ -68,7 +68,7 @@ public struct Region: RawRepresentable, Equatable {
     public static var mecentral1: Region { .init(rawValue: "me-central-1") }
     // Middle East (Bahrain)
     public static var mesouth1: Region { .init(rawValue: "me-south-1") }
-    // South America (São Paulo)
+    // South America (Sao Paulo)
```
0xTim, line 71: OOI - any reason for this?

0xTim (approved):

Sources/SotoCore/Doc/Region.swift
adam-fowler: Region.swift is a generated file. That'll come from the endpoints.json file from AWS.

PR #539, opened 2023-03-26
Add Concurrency primitives, functionality

adam-fowler:
- Added `AsyncSequence.enumerated()` which does the same as `Sequence.enumerated()`.
- Added `FixedSizeByteBufferAsyncSequence` which is constructed from an `AsyncSequence` that outputs arbitrary sized ByteBuffers and instead outputs ByteBuffers of a fixed size.
- Added `Sequence.concurrentMap`which runs an async function concurrently on all the members of a sequence while limiting the number of concurrent tasks that can run.

None of these are actually used by SotoCore yet although they will be in the future. But they are all used in the new s3 multipart upload from `AsyncSequence` implementation. https://github.com/soto-project/soto/pull/664
 
At the same time I have moved async code specific to certain types out of AsyncAwaitSupport folder and next to the relevant file.


ktoso:
open an issue about the concurrentMap; there's a long story for why we don't provide it -- the decision "what should the width be" has so far prevented shipping such thing, but we very much could. Open a ticket please (on apple/swift)

ktoso:
Alternatively it could live in async algorithms

adam-fowler:
The one I was more surprised about was there is no version of `AsyncSequence.enumerated()`.

ktoso:
That's an easy one to propose to async algorithms, I'm sure it'd be welcomed there :-)

PR #542, opened 2023-04-11
Use product platform condition for swift-crypto

adam-fowler:
Get rid of `if` statement. This should mean we can now do cross compilation.


0xTim (approved):

adam-fowler:
Looks like this failed on macOS. Frustratingly I couldn't replicate this failure.

0xTim:
Definitely on the right branch? I can reproduce it

adam-fowler:
I think it sometimes caches stuff and misses that issue. Looks like I'm going to have to park this PR until I force availability to macOS 10.15 etc

PR #544, opened 2023-06-06
Add preconcurrency flag correctly on Linux

adam-fowler:


adam-fowler:
@gwynne

gwynne:
@adam-fowler IIRC, `@preconcurrency` ended up being broken on Swift 5.6, such that in NIO it's only used when `#if swift(>=5.7)` (ex.: https://github.com/apple/swift-nio/blob/main/Sources/NIOCore/EventLoopFuture.swift#L541-L601); is it deliberate that you're doing the exact opposite on non-Linux? 😕

adam-fowler:
> @adam-fowler IIRC, `@preconcurrency` ended up being broken on Swift 5.6, such that in NIO it's only used when `#if swift(>=5.7)` (ex.: https://github.com/apple/swift-nio/blob/main/Sources/NIOCore/EventLoopFuture.swift#L541-L601); is it deliberate that you're doing the exact opposite on non-Linux? 😕

I thought `@preconcurrency` was fine for imports on 5.6. Anyway `Date` and `Data` were marked `Sendable` on macOS swift 5.7 and later which is why the check is the opposite way

Tests/SotoCoreTests/AWSRequestTests.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if compiler(<5.7)
+#if compiler(<5.7) || os(Linux)
```
gwynne, line 15: Looks like you missed changing the `#if`s here and in `DictionaryEncoderTests.swift`, if this isn't on purpose

adam-fowler: Yep looks like I missed that one

adam-fowler: fixed

0xTim (approved):

PR #549, opened 2023-06-17
Remove eventLoop parameter from async execute APIs

adam-fowler:


Sources/SotoCore/AWSClient+async.swift
```diff
@@ -332,6 +316,86 @@ extension AWSClient {
...
+                    try await Task.sleep(nanoseconds: UInt64(retryTime.nanoseconds))
+                } else {
+                    throw error
+                }
+            }
+            attempt += 1
+        }
+    }
+
+    /* func execute(attempt: Int) {
```
0xTim, line 359: Should this function be removed?

0xTim (approved):

Sources/SotoCore/AWSClient+async.swift
adam-fowler: removed

PR #552, opened 2023-06-17
Remove EventLoop version of Endpoint Discovery code

adam-fowler:
This ended up being slightly more complex than the removal of the waiter and paginators EL code as I had to rewrite the endpoint discovery in Swift Concurrency.

 I have added a new type `ExpiringValue` which stores a value and an expiration date for that value. When you get the value you have to provide a function that would update the value if it is out of date. The type also ensures you aren't calling the update function while an update is already in progress (instead you wait on the result of the in progress update). This will be used in other parts of the codebase when I get to them eg some of the credential providers.


Sources/SotoCore/AWSEndpointDiscovery.swift
```diff
@@ -39,96 +39,52 @@ public struct AWSEndpoints {
...
+public struct AWSEndpointStorage: Sendable {
+    let endpoint: ExpiringValue<String>
 
-    /// Initialize endpoint storage
-    /// - Parameter endpoint: Initial endpoint to use
-    public init(endpoint: String) {
-        self.expiration = Date.distantPast
-        self.endpoint = endpoint
+    public init() {
+        self.endpoint = .init(threshold: 3 * 60)
```
0xTim, line 47: Is there a reason why this is hard coded as 180?

Sources/SotoCore/Concurrency/ExpiringValue.swift
```diff
@@ -0,0 +1,96 @@
...
+import Foundation
+import Logging
+
+/// Type holding a value and an expiration value.
+///
+/// When accessing the value you have to provide a closure that will update the
+/// value if it has expired or is about to expire. The type ensures there is only
+/// ever one value update running at any one time. If an update is already running
+/// when you call `getValue` it will wait on the current update function to finish.
+actor ExpiringValue<T> {
```
0xTim, line 24: Does this need to be public?

adam-fowler: Not at this point. There is a possibility I will use it in some Cognito credential providers. But for the moment I am only using it internally

Sources/SotoCore/AWSEndpointDiscovery.swift
adam-fowler: It is set to 3 minutes. Basically when an endpoint is about to expire it has 3 minutes to get a new one. I didn't think this was something that needed a configurable value. I've added a comment

0xTim (approved):

PR #555, opened 2023-06-20
Use Async calls from AsyncHTTPClient

adam-fowler:
This was complex. Trying to return back the AsyncSequence from AHC in a generic way until I found `AnyAsyncSequence` in AHC. This simplified passing the body back from AHC through my own types.
- Added `AnyAsyncSequence` as a generic method for passing AsyncSequences about.
- Added `ByteBufferAsyncSequence` as a way to convert a `ByteBuffer` into an `AsyncSequence` of `ByteBuffer` slices.
- Added `HTTPBody` to hold response payloads. This'll probably be also used in requests but I haven't got to that yet. If so I'll rename back to `AWSPayload` for some consistency. The `HTTPBody` no longer defines the type of data it holds (eg xml, json, raw data) this data is got from the serviceProtocol

Streaming requests and response currently don't work, so related code and tests may be commented out


Tests/SotoCoreTests/AWSClientTests.swift
```diff
@@ -261,217 +261,217 @@ class AWSClientTests: XCTestCase {
...
-                let response = AWSTestServer.Response(httpStatus: .ok, headers: [:], body: nil)
-                return .result(response)
-            }
-
-            try await responseTask
-        } catch {
-            XCTFail("Unexpected error: \(error)")
-        }
-    }
+    /*    func testRequestStreaming() async throws {
```
0xTim, line 264: Is this to be added back in or deleted?

Tests/SotoCoreTests/AWSClientTests.swift
```diff
@@ -746,101 +746,101 @@ class AWSClientTests: XCTestCase {
...
-            var byteBuffer = ByteBufferAllocator().buffer(capacity: bufferSize)
-            byteBuffer.writeBytes(data)
-            let response = AWSTestServer.Response(httpStatus: .ok, headers: ["test": "TestHeader"], body: byteBuffer)
-            return .result(response)
-        })
-
-        _ = try await responseTask
-        XCTAssertEqual(countAtomic.load(ordering: .relaxed), 0)
-    }
+    /* func testStreamingResponse() async {
```
0xTim, line 749: As above

Sources/SotoCore/HTTP/AsyncHTTPClient.swift
```diff
@@ -27,103 +27,41 @@ extension AsyncHTTPClient.HTTPClient {
...
-        timeout: TimeAmount,
-        on eventLoop: EventLoop? = nil,
+        on eventLoop: EventLoop,
         logger: Logger,
         stream: @escaping AWSResponseStream
-    ) async throws -> AWSHTTPResponse {
-        let eventLoop = eventLoop ?? self.eventLoopGroup.any()
-        return try await self.execute(request: request, timeout: timeout, on: eventLoop, logger: logger, stream: stream).get()
+    ) async throws -> HTTPClientResponse {
+        preconditionFailure("Not supported")
```
0xTim, line 65: Should we just remove this if it's not supported?

Sources/SotoCore/HTTP/AsyncHTTPClient.swift
```diff
@@ -27,103 +27,41 @@ extension AsyncHTTPClient.HTTPClient {
...
-            requestBody = .stream(length: reader.contentSize) { writer in
-                return writer.write(reader: reader, on: eventLoop)
-            }
-        case .empty:
+            requestBody = .bytes(byteBuffer)
+        /* case .asyncSequence(let sequence, let length):
+         requestBody = .stream(
+             sequence,
+             length: length.map { .known($0) } ?? .unknown
+         ) */
```
0xTim, line 41: Commented out code

Tests/SotoCoreTests/AWSClientTests.swift
adam-fowler: I should have made it clear streaming requests/responses don't work yet. They will the subject of the next few PRs. So I commented out the tests that would fail and will uncomment them as features are implemented.

Tests/SotoCoreTests/AWSClientTests.swift
adam-fowler: see above

Sources/SotoCore/HTTP/AsyncHTTPClient.swift
adam-fowler: It will get removed in the next PR. Currently it a stub used by another section of code

Sources/SotoCore/HTTP/AsyncHTTPClient.swift
adam-fowler: This is commented out while streaming requests don't work. It will be re-instated when I get them working

adam-fowler:
Fixed the tests. There was a weird interaction between the test server which can only receive one request at a time and me not flush the response body when I don't expect one

PR #558, opened 2023-06-22
7.x.x: Response streaming

adam-fowler:
Replace old streaming functions that call a closure with an AsyncSequence (currently HTTPBody) added to the output type.

### DIFF
--- a/Sources/SotoCore/Message/AWSResponse.swift
+++ b/Sources/SotoCore/Message/AWSResponse.swift
@@ -32,8 +32,8 @@ public struct AWSResponse {
     /// initialize an AWSResponse Object
     /// - parameters:
     ///     - from: Raw HTTP Response
-    ///     - serviceProtocol: protocol of service (.json, .xml, .query etc)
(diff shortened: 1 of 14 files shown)


Sources/SotoCore/Message/AWSResponse.swift
```diff
@@ -27,50 +27,25 @@ public struct AWSResponse {
...
     public var headers: HTTPHeaders
     /// response body
-    public var body: Body
+    public var body: HTTPBody
 
     /// initialize an AWSResponse Object
     /// - parameters:
     ///     - from: Raw HTTP Response
     ///     - serviceProtocol: protocol of service (.json, .xml, .query etc)
     ///     - raw: Whether Body should be treated as raw data
```
0xTim, line 36: These need updating

0xTim (approved):

Sources/SotoCore/Message/AWSResponse.swift
adam-fowler: done

PR #560, opened 2023-06-22
7.x.x: S3 Signed request streaming

adam-fowler:
I was shocked how easy this was to write after the pain I went through to write the original with EventLoops. It is just an AsyncSequence that wraps the original data stream AsyncSequence. Internally when I create the AsyncIterator I wrap the sequence in a `FixedSizeByteBufferAsyncSequence` so I can get a sequence of fixed sized ByteBuffers and then my iterator signs each of these individually.

This PR also a number of minor fixes I found while building Soto service files
- HTTPBody(asyncSequence:length) is now public
- Added a helper init function for `AWSResponseError`
- ByteBufferAsyncSequence is public (we removed the version from Soto)


Sources/SotoCore/AWSClient.swift
```diff
@@ -521,7 +521,7 @@ extension AWSClient {
         url: URL,
         httpMethod: HTTPMethod,
         headers: HTTPHeaders = HTTPHeaders(),
-        body: AWSPayload,
+        body: HTTPBody,
```
0xTim, line 524: Does the DocC comment need updating?

0xTim (approved):

Sources/SotoCore/AWSClient.swift
adam-fowler: No payload/body are basically same thing. Did you see my slack comment about renaming HTTPBody to AWSPayload?

PR #563, opened 2023-07-06
Fixes for when strict-concurrency is on

adam-fowler:


Package.swift
```diff
@@ -57,7 +57,8 @@ let package = Package(
                 .product(name: "NIOTransportServices", package: "swift-nio-transport-services"),
                 .product(name: "NIOFoundationCompat", package: "swift-nio"),
                 .product(name: "JMESPath", package: "jmespath.swift"),
-            ]
+            ],
+            swiftSettings: [.unsafeFlags(["-Xfrontend", "-warn-concurrency"])]
```
0xTim, line 61: Is this to make it easy to turn strict on? I think warn is the default right?

0xTim (approved):

Package.swift
adam-fowler: How are you turning strict on. This is how I thought it was done.

PR #564, opened 2023-07-10
Rename HTTPBody to AWSHTTPBody

adam-fowler:


adam-fowler:
This got merged with #565

PR #566, opened 2023-07-19
Add support for decoding event streams

adam-fowler:
See S3.SelectObjectContent, Lambda.InvokeWithResponseStream etc. You can see documentation on decoding these here https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html.

- Added an `AWSEventStream` type that reads from a base ByteBuffer async sequence.
- These buffers are passed to an EventStreamDecoder which decodes the headers and then passes payload onto relevant decoder.

Related PRs: https://github.com/soto-project/soto-codegenerator/pull/74, https://github.com/soto-project/soto/pull/686

### DIFF
--- a/Sources/SotoCore/Encoder/EventStreamDecoder.swift
+++ b/Sources/SotoCore/Encoder/EventStreamDecoder.swift
@@ -1,0 +1,218 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Soto for AWS open source project
+//
+// Copyright (c) 2023 the Soto project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Soto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import NIOCore
+import SotoXML
+
+/// Event stream decoder. Decodes top level `:event-type` header and then passes the payload
+/// to another decoder based off the `:content-type` header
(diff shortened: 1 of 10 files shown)


Sources/SotoCore/Encoder/EventStreamDecoder.swift
```diff
@@ -0,0 +1,218 @@
...
+    func singleValueContainer() throws -> SingleValueDecodingContainer {
+        preconditionFailure("Single value containers are not supported")
+    }
+}
+
+/// Container used for passed event payload to decoders
+public struct EventDecodingContainer {
+    let payload: ByteBuffer
+
+    public func decodePayload() -> ByteBuffer {
```
0xTim, line 193: Minor nit - this has no DocC comments

Sources/SotoCore/Encoder/EventStreamDecoder.swift
```diff
@@ -0,0 +1,218 @@
...
+import Foundation
+import NIOCore
+import SotoXML
+
+/// Event stream decoder. Decodes top level `:event-type` header and then passes the payload
+/// to another decoder based off the `:content-type` header
+struct EventStreamDecoder {
+    public init() {}
+
+    public func decode<T: Decodable>(_ type: T.Type, from buffer: inout ByteBuffer) throws -> T {
```
0xTim, line 24: Do these need to be public if the type isn't?

0xTim (approved):
Minor nits about access control and doc comments but take or leave them

PR #567, opened 2023-07-19
Reduce exports from SotoSignerV4

adam-fowler:


0xTim (approved):
Reducing exports is always good - is there any reason why these are needed at all for v7?

adam-fowler:
> Reducing exports is always good - is there any reason why these are needed at all for v7?

I think it is a good idea to export the symbols my public APIs use. Makes it easier for the end user.

0xTim:
Oh completely agree - are things like ByteBuffer going to be part of the public API though?

(This is a more general question as opposed to being specifically about this PR, I know removing all the future APIs is a multi step process)

adam-fowler:
ByteBuffer is part of the public API. I don't think that is going to change. Unless we get the bag of bytes type in the sodlib

0xTim:
🤦‍♂️ travel brain getting futures and byte buffers mixed up

PR #568, opened 2023-07-20
Simplify CodeQL file now it's out of Beta

0xTim:
This removes the environment variable as Swift is out of beta so no longer needed. It also removes some permissions that are only required for private repositories.

We could also run this on Linux as CodeQL now supports 5.8 and Linux properly, but I've left it as macOS. Happy to switch it to Linux if it makes it easier (currently in Vapor I've opted for both to catch any issues just in case)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -6,16 +6,11 @@ on:
   pull_request:
     branches: [ "main" ]
 
-env:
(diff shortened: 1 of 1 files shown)


adam-fowler:
That didn't seem to work

0xTim:
Hmm this is weird, it was the `swift build` command that just failed. I've noticed slower bios times on the Vapor side with CodeQL enabled, I might have to raise that in Slack

adam-fowler:
It seems happy now

PR #569, opened 2023-07-21
Merge request and response types

adam-fowler:
AWSRequest and AWSHTTPRequest have been merged to AWSHTTPRequest
AWSResponse and AWSHTTPResponse have been merged to AWSHTTPResponse

As request doesn't include some information it had before which was used in middleware this has been moved to the middleware context.


Sources/SotoCore/AWSClient.swift
```diff
@@ -553,26 +560,37 @@ extension AWSClient {
...
-            .applyMiddlewares(serviceConfig.middlewares + middlewares, config: serviceConfig)
+        var response = response
+        if !raw {
+            try await response.collateBody()
+        }
+        return try response.generateOutputShape(operation: operationName, serviceProtocol: serviceConfig.serviceProtocol)
+    }
 
-        return try awsResponse.generateOutputShape(operation: operationName, serviceProtocol: serviceConfig.serviceProtocol)
+    /// Generate an AWS Response from  the operation HTTP response and return the output shape from it. This is only every called if the response includes a successful http status code
```
0xTim, line 577: Very minor nit
```suggestion
    /// Generate an AWS Response from the operation HTTP response and return the output shape from it. This is only ever called if the response includes a successful http status code
```

Sources/SotoCore/HTTP/AWSHTTPResponse.swift
```diff
@@ -26,39 +26,31 @@ import NIOFoundationCompat
...
-            self.body = try await .init(buffer: response.body.collect(upTo: .max))
-        }
+    /// Initialize AWSResponse
+    init(status: HTTPResponseStatus, headers: HTTPHeaders, body: AWSHTTPBody = .init()) {
+        self.status = status
+        self.headers = headers
+        self.body = body
+    }
+
+    /// collate body
```
0xTim, line 45: Given this isn't part of the public API this comment seems a bit redundant 😅

0xTim (approved):
LGTM - couple of nits

Tests/SotoCoreTests/AWSClientTests.swift
```diff
@@ -576,52 +576,6 @@ class AWSClientTests: XCTestCase {
         }
     }
 
-    /// verify we are not calling the Retry handler when streaming a request
-    func testDontRetryStreamingRequests() async {
```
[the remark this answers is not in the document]
adam-fowler: Yes we don't need it anymore. Previously I couldn't run retry on streamed responses because they called a callback for each data chunk. If you retried the request you could be sending the same data twice. Now we return the header and an async sequence with the body data and this is no longer a concern.

You can see in AWSClient.swift line 451 the related code has been deleted

PR #570, opened 2023-07-24
Middleware restructure

adam-fowler:
- Rewrite middleware to pass request onto next middleware instead of just editing request/response and returning it. eg 
```swift
struct MyMiddleware: AWSMiddlewareProtocol {
    public func handle(
        _ request: AWSHTTPRequest, 
        context: AWSMiddlewareContext, 
        next: (AWSHTTPRequest, AWSMiddlewareContext) async throws -> AWSHTTPResponse
    ) async throws -> AWSHTTPResponse {
        let request = doStuffToRequest(request)
        let response = try await next(request, context)
        return doStuffToResponse(response)
    }
}
```
- Use result builder `AWSMiddlewareStack` to build middleware stack
- Convert each stage of the Soto execute chain to a middleware eg signing, retry, error handling, endpoint discovery
- execute is now 
  - create request, 
  - run middleware on request, passing final request to http client 
  - process response returned from middleware to create output object


0xTim (approved):
Some minor documentation nits but LGTM

PR #581, opened 2023-08-11
Soto 7.0.0-alpha.1: fix Package.swift versions

MahdiBM:
The low `swift-crypto` version is problematic.
I'm not sure if you actually meant that 🤔.

Also there is `.package(url: "https://github.com/apple/swift-metrics.git", "1.0.0"..<"3.0.0"),` which seems suspicious to me, so just want to point it out if you actually meant `"1.0.0"..<"3.0.0"`.
I have no idea, but thought doesn't hurt to mention it.

Screenshot 2023-08-11 at 9 02 20 PM

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -30,7 +30,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0"),
-        .package(url: "https://github.com/apple/swift-crypto.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", from: "2.0.0"),
         .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.0.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
         .package(url: "https://github.com/apple/swift-metrics.git", "1.0.0"..<"3.0.0"),
--- a/Package.swift
+++ b/Package.swift
@@ -30,10 +30,10 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0"),
-        .package(url: "https://github.com/apple/swift-crypto.git", from: "2.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", "1.0.0"..<"3.0.0"),
         .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.0.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
-        .package(url: "https://github.com/apple/swift-metrics.git", "2.0.0"),
+        .package(url: "https://github.com/apple/swift-metrics.git", "1.0.0"..<"3.0.0"),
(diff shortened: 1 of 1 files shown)


gwynne:
Looks like a transposal typo - "1.0.0" ..< "3.0.0" was previously the recommended usage for `swift-crypto` (although I think that's obsolete at this point).

MahdiBM:
yeah i was suspicious of that basically.
Though the tracing package is at 2.0.0+, so not sure how 1.0.0 ended up in swift-crypto.

Package.swift
```diff
@@ -30,10 +30,10 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0"),
-        .package(url: "https://github.com/apple/swift-crypto.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", from: "2.0.0"),
         .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.0.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
-        .package(url: "https://github.com/apple/swift-metrics.git", "1.0.0"..<"3.0.0"),
```
0xTim, line 36: This is fine

Package.swift
```diff
@@ -30,10 +30,10 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0"),
-        .package(url: "https://github.com/apple/swift-crypto.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", from: "2.0.0"),
```
0xTim, line 33: This can be 1.0.0 to 3.0.0

0xTim (changes requested):

Package.swift
adam-fowler: Yes this was the recommendation. Swift-metrics had to make a breaking change to fix something. But this meant the major version being used was quite varied across the eco-system. So the suggestion was to allow version 1.0.0 ..< 3.00.

gwynne: It can, but shouldn't. Anything that's still referring to SwiftCrypto 1.x without allowing the 2.x range is now so badly out of date that it _should_ be forced to update; being compatible with such out of date dependencies is a bad idea.

adam-fowler: We could do the same for swift-crypto as @0xTim suggests. I don't require v2.0 so why should I force it on my users.

gwynne: Not to mention that given the 5.6 minimum Swift requirement, this package effectively requires SwiftCrypto >= 2.5.0

adam-fowler: The recommendation from Cory when he released 2.0.0 was to support the 1.0.0 ..< 3.0.0 range. Given Soto only uses MD5, SHA1 and SHA256 don't think requiring 2.0.0 is necessary.

MahdiBM:
better now? :eyes:

0xTim (approved):

adam-fowler (approved):

adam-fowler:
And released https://github.com/soto-project/soto-core/releases/tag/7.0.0-alpha.2

PR #603, opened 2024-03-05
Add Sendable to AWSShapeOptions

sidepelican:
In Xcode15.3(now in RC), it shows many warnings of concurrency.


adam-fowler (approved):
Most of the Soto Sendable issues are fixed in the 7.0.0-alpha. I wasn't intending on fixing all of them for 6.x.x but there is no harm in merging this though.

PR #617, opened 2024-09-19
Swift 6

adam-fowler:


Tests/SotoCoreTests/AWSClientTests.swift
```diff
@@ -52,42 +52,43 @@ class AWSClientTests: XCTestCase {
...
+                serviceProtocol: .json(version: "1.1"),
+                endpoint: awsServer.address
+            )
+            let client = createAWSClient(
+                credentialProvider: .static(accessKeyId: "foo", secretAccessKey: "bar"),
+                httpClient: httpClient
+            )
+            defer {
+                XCTAssertNoThrow(try client.syncShutdown())
+            }
```
0xTim, line 66: Would it make sense to move this out so put the client inside the teardown with the others?

0xTim (approved):

Tests/SotoCoreTests/AWSClientTests.swift
adam-fowler: done

PR #627, opened 2024-12-11
Add support for S3 arn based access points

adam-fowler:
You can set up an accesspoint using S3Control.
This PR adds support for using this arn as a bucket name.

This has been tested with S3 access points
In theory this should also work for S3 outposts and S3 object lambdas (thus resolving https://github.com/soto-project/soto/issues/611).

### DIFF
--- a/Sources/SotoCore/Doc/ARN.swift
+++ b/Sources/SotoCore/Doc/ARN.swift
@@ -13,6 +13,11 @@
 //===----------------------------------------------------------------------===//
 
 /// Amazon Resource Name (ARN). A unique identifier assigned to AWS resource
+///
+/// Comes in one of the following forms
+/// - arn:partition:service:region:account-id:resource-id
+/// - arn:partition:service:region:account-id:resource-type/resource-id
+/// - arn:partition:service:region:account-id:resource-type:resource-id
 public struct ARN {
     public init?<S: StringProtocol>(string: S) where S.SubSequence == Substring {
         let split = string.split(separator: ":", omittingEmptySubsequences: false)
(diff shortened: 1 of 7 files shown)


Sources/SotoCore/Middleware/Middleware/S3Middleware.swift
```diff
@@ -37,87 +37,173 @@ internal import SotoXML
...
-                    // Bucket name is part of host. No need to append bucket
-                    urlHost = host
-                } else {
-                    urlHost = "\(bucket).\(host)"
-                }
-            } else {
-                urlPath = paths.joined(separator: "/")
+        } else if isAmazonUrl || context.serviceConfig.options.contains(.s3ForceVirtualHost), !bucket.contains(".") {
+            let pathsWithoutBucket = paths.dropFirst()  // bucket
+            urlPath = pathsWithoutBucket.first.flatMap { String($0) } ?? ""  //pathsWithoutBucket.joined(separator: "/")
```
0xTim, line 114: Looks like some dead code here

Sources/SotoCore/Doc/ARN.swift
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Soto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public struct ARN {
```
0xTim, line 15: Should we add some docs here to explain the type and also for how the init breaks it down

0xTim (approved):
Couple of comments otherwise LGTM

PR #632, opened 2025-01-01
Fix encoding of dates in URL queries

adam-fowler:


petrpavlik (approved):
Thanks! This should solve my issue.

0xTim (approved):

Issue #633, opened 2025-01-25
Build error on nightlies


MahdiBM:
soto-core no longer builds on Swift 6.1 / 6.2 (main) nightlies.

There is a compilation error complaining about [this line](https://github.com/soto-project/soto-core/blob/d2b2726eb7ef7f054b64264cf2e1884a3182e4a9/Sources/SotoCore/Encoder/CodableProperties/DateCoders.swift#L112) not being concurrency safe.

```swift
public static let dateFormatter = createDateFormatter()
```

Perhaps it should use `@nonisolated(unsafe)` like the other formatters above it are doing.

MahdiBM:
@adam-fowler can you try a macOS nightly toolchain as well? That's what I'm using.

adam-fowler:
> [@adam-fowler](https://github.com/adam-fowler) can you try a macOS nightly toolchain as well? That's what I'm using.

Is that main or 6.1? Also what date?

MahdiBM:
Doesn't matter just the latest.
IIRC 10th January main, or 23rd 6.1.

adam-fowler:
A PR would be much appreciated. Not finding much time at the moment.

vsarunas:
@adam-fowler , I have opened up https://github.com/soto-project/soto-core/pull/639 as running into this as well.

PR #639, opened 2025-02-12
Build on Swift 6.1 under macOS Static property 'dateFormatter' is not concurrency-safe

vsarunas:
Avoid  build error on Swift 6.1 under macOS:

```
soto-core/Sources/SotoCore/Encoder/CodableProperties/DateCoders.swift:112:23 Static property 'dateFormatter' is not concurrency-safe because non-'Sendable' type 'DateFormatter' may have shared mutable state
```

Linux compiles fine.

Fixes https://github.com/soto-project/soto-core/issues/633


adam-fowler:
thanks

PR #642, opened 2025-02-27
Support Exception Events

grahamburgsma:
This adds support for decoding `exception` events from an event stream. 

Example stream with exception events: [bedrockagentruntime#ResponseStream](https://github.com/aws/aws-sdk-go-v2/blob/f0d143d1fb449b320ac5ff0b2c513812a51d9904/codegen/sdk-codegen/aws-models/bedrock-agent-runtime.json#L7491)

### DIFF
--- a/Sources/SotoCore/Encoder/EventStreamDecoder.swift
+++ b/Sources/SotoCore/Encoder/EventStreamDecoder.swift
@@ -211,15 +211,51 @@ extension CodingUserInfoKey {
 }
 
 /// Errors thrown while decoding the event stream buffers
-enum AWSEventStreamError: Error {
+public struct AWSEventStreamError: Error {
+    public struct Code: Sendable {
+        enum _Internal {
+            case corruptHeader
+            case missingHeader
+            case corruptPayload
+            case errorMessage
+            case unsupportedContentType
+        }
+
+        private let value: _Internal
+
+        /// The message headers are corrupt
+        public static var corruptHeader: Self { .init(value: .corruptHeader) }
+        /// An event stream message headers is missing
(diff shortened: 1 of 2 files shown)


adam-fowler (commented):
Jeez bedrock is a real trouble maker
Is there any documentation on the `:exception-type` header?

grahamburgsma:
@adam-fowler 

> Jeez bedrock is a real trouble maker

Sure is! It has been trouble in many ways, not fun to work with.

> Is there any documentation on the :exception-type header?

Best resource I was found was this https://sdk.amazonaws.com/kotlin/api/smithy-kotlin/api/1.4.3/aws-event-stream/aws.smithy.kotlin.runtime.awsprotocol.eventstream/-message-type/-exception/index.html

adam-fowler:
Currently message processing is spread across two sites. Are you up for tidying this up or do you want to merge this and I'll do that. Basically I would prefer
```swift
switch self.headers[":message-type"] {
case "event"
    get required headers else { throw `AWSEventStreamError.missingHeader` } 
case "exception"
    get required headers else { throw `AWSEventStreamError.missingHeader` }
case "error"
    get required headers else { throw `AWSEventStreamError.missingHeader` }
    `AWSEventStreamError.errorMessage` 
}
```
- remove the line checking for `:message-type` in readEvent
- You'll need to add a missingHeader error to `AWSEventStreamError`. Then I've seen the AWSEventStreamError isn't public 🤦 . Actually don't worry about that, I'll fix that one.

adam-fowler:
Go ahead. It was waiting for a review, but you can merge it.

adam-fowler (approved):
Looking good, need to update to latest from main before I can merge though

Issue #532, opened 2022-12-27
Support SigV4a


adam-fowler:

MashfiqueA:
Any movement or updates on this? We started using S3 Multi-Region Access Point (MRAP) for our app and having this under soto would be awesome.

Also, is there any alternative available for iOS/Swift?

adam-fowler:
No not really. It's a big task with a number of unknowns.

fabianfett:
I'm working on this. Making good progress!

MashfiqueA:
> I'm working on this. Making good progress!

@fabianfett happy to help with testing and code review!

fabianfett:
Opened the pr:
https://github.com/soto-project/soto-core/pull/645

PR #645, opened 2025-03-20
Add support for sigV4a signing

fabianfett:
- Tested signing manually against S3 for both query and header
- Writing a full test case is hard, as signing ECDSA signing is non deterministic
- Added new HexEncoding mode, that can be used to make `hexDigest()` much faster :)

Fixes: #532

### DIFF
--- a/Sources/SotoSignerV4/SigV4a.swift
+++ b/Sources/SotoSignerV4/SigV4a.swift
@@ -13,7 +13,11 @@
 //===----------------------------------------------------------------------===//
 
 import Crypto
-import Foundation
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import struct Foundation.Data
+#endif
 
 package struct SigV4aKeyPair {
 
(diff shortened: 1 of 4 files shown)


Sources/SotoSignerV4/SigV4a.swift
```diff
@@ -0,0 +1,248 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Soto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crypto
+import Foundation
```
adam-fowler, line 16: Given there is some effort to replace Foundation with FoundationEssentials where possible can you do this here

Sources/SotoSignerV4/SigV4a.swift
```diff
@@ -0,0 +1,248 @@
...
+    package static func compareConstantTime(lhs: [UInt8], rhs: [UInt8]) -> Int8 {
+        guard lhs.count == rhs.count else {
+            fatalError("Input arrays must be of same size")
+        }
+
+        var gt: UInt8 = 0
+        var eq: UInt8 = 1
+        let length = lhs.count
+
+        for i in 0..<length {
```
adam-fowler, line 99: Can you a comment explaining what is happening here

Sources/SotoSignerV4/SigV4a.swift
```diff
@@ -0,0 +1,248 @@
...
+
+package struct HexEncoding<Base: Sequence> where Base.Element == UInt8 {
+    var base: Base
+
+    package init(_ base: Base) {
+        self.base = base
+    }
+}
+
+extension HexEncoding: Sequence {
```
adam-fowler, line 161: Can you move this into a separate file `HexEncoding.swift`

Sources/SotoSignerV4/SigV4a.swift
```diff
@@ -0,0 +1,248 @@
...
+            return .nextCounter
+        }
+
+        return try .success(P256.Signing.PrivateKey(rawRepresentation: bytes.addingOne()))
+    }
+}
+
+extension [UInt8] {
+
+    package mutating func addOne() {
```
adam-fowler, line 136: Does this need to be `package`

Sources/SotoSignerV4/SigV4a.swift
```diff
@@ -0,0 +1,248 @@
...
+        var carry: UInt32 = 1
+        for index in self.indices.reversed() {
+            var digit = UInt32(self[index])
+            digit += carry
```
adam-fowler, line 140: Once carry is 0 can't you just break out of the loop

Sources/SotoSignerV4/SigV4a.swift
```diff
@@ -0,0 +1,248 @@
...
+        var base: Base.Index
+        var first: Bool
+    }
+
+    package var startIndex: Index {
+        Index(base: self.base.startIndex, first: true)
+    }
+
+    package var endIndex: Index {
+        Index(base: self.base.endIndex, first: true)
```
adam-fowler, line 221: Should `first` be set here?

adam-fowler (commented):
In general looks pretty good. Couple of minor issues
One question is there a plan to provide some API to choose sigv4a in AWSClient

Sources/SotoSignerV4/SigV4a.swift
fabianfett: yes. endIndex is the first index that is out of bounds. given that we start with first = true, this is correct.

PR #650, opened 2025-04-15
Don't output body for json DELETE requests

adam-fowler:
HTTP DELETE methods should not include a body. We were only checking for GET and HEAD method requests previously. It wasn't an issue but AWS lambda is now returning a signing error because of this. So I've added a check for DELETE method in the serialisation of json bodies.

EDIT: Verified it fixes signing issue with deleting lambdas

### DIFF
--- a/Sources/SotoCore/HTTP/AWSHTTPRequest.swift
+++ b/Sources/SotoCore/HTTP/AWSHTTPRequest.swift
@@ -128,7 +128,8 @@ extension AWSHTTPRequest {
             encoder.userInfo[.awsRequest] = requestEncoderContainer
             encoder.dateEncodingStrategy = .secondsSince1970
             let buffer = try encoder.encodeAsByteBuffer(input, allocator: configuration.byteBufferAllocator)
-            if method == .GET || method == .HEAD, buffer == ByteBuffer(string: "{}") {
+            // GET, HEAD and DELETE methods should have an empty body
+            if method == .GET || method == .HEAD || method == .DELETE, buffer == ByteBuffer(string: "{}") {
                 body = .init()
             } else {
                 body = .init(buffer: buffer)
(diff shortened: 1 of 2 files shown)


Sources/SotoCore/HTTP/AWSHTTPRequest.swift
```diff
@@ -260,6 +261,8 @@ extension AWSHTTPRequest {
         guard self.headers["content-type"].first == nil else {
             return
         }
+        // in theory we don't need this check as the check for an empty body should skip adding
+        // a header and a GET or HEAD request should have an empty body
         guard self.method != .GET, self.method != .HEAD else {
```
0xTim, line 266: Do we not want to check for `DELTE` here as well?

Tests/SotoCoreTests/AWSRequestTests.swift
```diff
@@ -501,13 +501,23 @@ class AWSRequestTests: XCTestCase {
...
-        var request: AWSHTTPRequest?
-        XCTAssertNoThrow(request = try AWSHTTPRequest(operation: "Test", path: "/", method: .POST, input: input, configuration: config))
-        XCTAssertEqual(request?.body.asString(), "{}")
+        let request = try AWSHTTPRequest(operation: "Test", path: "/", method: .POST, input: input, configuration: config)
+        XCTAssertEqual(request.body.asString(), "{}")
+        XCTAssertEqual(request.headers["content-type"].first, "application/x-amz-json-1.0")
+    }
+
+    /// JSON POST request require a body even if there is no data to POST
+    func testEmptyGetJsonObject() throws {
```
0xTim, line 514: Worth adding a delete test?

0xTim (approved):

Tests/SotoCoreTests/AWSRequestTests.swift
adam-fowler: added

0xTim (approved):

PR #652, opened 2025-05-02
Valid credential threshold

adam-fowler:
Add a threshold to expiration for when RotatingCredentialProvider returns valid credentials.
Previous there was one threshold. When you were within this threshold of credentials expiring, it would request new credentials, but still use the old ones as they were still valid. This has one issue though, the credential provider can return credentials that will expire in the next millisecond as valid.

This PR adds a second threshold. If within this threshold then the credential provider will wait until new credentials are available, instead of return the old credentials. By default this has been set to 15 seconds, but you can override it in the initializer for RotatingCredentialProvider.

### DIFF
--- a/Sources/SotoCore/Credential/RotatingCredentialProvider.swift
+++ b/Sources/SotoCore/Credential/RotatingCredentialProvider.swift
@@ -24,17 +24,46 @@ import struct Foundation.TimeInterval
 ///
 /// Used for wrapping another credential provider whose `getCredential` method returns an `ExpiringCredential`.
 /// If no credential is available, or the current credentials are going to expire in the near future  the wrapped credential provider
-/// `getCredential` is called again. If current credentials have not expired they are returned otherwise we wait on new
+/// `getCredential` is called again. If current credentials have not expired (within a threshold) they are returned otherwise we wait on new
 /// credentials being provided.
 public final class RotatingCredentialProvider: CredentialProvider {
     let expiringCredential: ExpiringValue<Credential>
-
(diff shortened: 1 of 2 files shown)


Sources/SotoCore/Credential/RotatingCredentialProvider.swift
```diff
@@ -24,17 +24,46 @@ import struct Foundation.TimeInterval
...
+    ///   - provider: Credential provider to request credentials from
+    ///   - remainingTokenLifetimeForUse: How near to expiration, before we request new credentials
+    public init(
+        context: CredentialProviderFactory.Context,
+        provider: CredentialProvider,
+        remainingTokenLifetimeForUse: TimeInterval? = nil
+    ) {
+        self.provider = provider
+        self.validCredentialThreshold = 15
+        self.expiringCredential = .init(threshold: remainingTokenLifetimeForUse ?? 165) {
```
0xTim, line 46: Why 165 here when the below is using 180? Would it be easier to call the new initialiser with the `validCredentialThreshold`?

0xTim (approved):

Sources/SotoCore/Credential/RotatingCredentialProvider.swift
adam-fowler: Previously it was 180 seconds. Now it is 165 seconds + 15 seconds from the validCredentialThreshold

0xTim: Ah yeah makes sense 👍

PR #665, opened 2025-07-08
Make AWSHTTPResponse initializer public.

### LINKED ISSUE #664: make AWSHTTPResponse initializer public
In order to make a custom implementation of `AWSHTTPClient`, I need to be able construct the protocol's return type, `AWSHTTPResponse`. Currently the initializer for this structure is not public, so one cannot be created outside of the `SotoCore` library.

jefflewis:
Fixes #664


adam-fowler (approved):
Looks good

PR #671, opened 2025-10-16
Add ServiceLifecycle support

adam-fowler:
Conform `AWSClient` to `Service`
Support is enabled with trait `ServiceLifecycleSupport` which is on by default

Also formatted everything to get rid of some validation errors


Sources/SotoCore/AWSClient+ServiceLifecycle.swift
```diff
@@ -0,0 +1,27 @@
...
+//===----------------------------------------------------------------------===//
+
+#if ServiceLifecycleSupport
+
+import ServiceLifecycle
+
+extension AWSClient: Service {
+    public func run() async throws {
+        // Ignore cancellation error
+        try? await gracefulShutdown()
```
0xTim: I find this API very confusingly named!

PR #686, opened 2026-02-18
feat. Add a SSOCredentialProvider

sebsto:
Add support for AWS IAM Identity Center (SSO) credentials. This lets soto-based apps authenticate using `aws sso login` cached tokens, the same way the AWS CLI does.         
                                                                                                                                                                   
Supports both the modern `sso-session` config format (with automatic token refresh via SSO-OIDC) and the legacy format (direct SSO fields in the profile).

The provider is added to the default credential chain, after `configFile()` and before `login()`.

To test this, you need to:
1. setup AWS IAM Identity Center in your AWS account (https://docs.aws.amazon.com/singlesignon/latest/userguide/getting-started.html)
2. configure the AWS CLI to authenticate with `aws sso login` (https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html) 

This implements https://github.com/soto-project/soto-core/issues/681

disclaimer : the swift docc comments and the majority of the tests have been generated automatically.

PR #690, opened 2026-05-19
Follow source_profile in the AWS config, and improve the handling of role_arn + SSO configurations

nicksloan:
This pull request improves support for AWS configuration files with the following changes:

- loadSharedCredentials now tolerates each file independently. Throws .noProvider only when both are unavailable; converts a downstream .missingProfile back to .noProvider if the credentials file was the missing one (preserving chain-fallthrough).
- parseSharedCredentials signature: Skips the parseCredentials call when credentialsINIParser is nil. Catches .missingProfile and proceeds when the config file supplies a role_arn; the static-credentials final path now explicitly requires a non-nil credentials.
- Assume-role branch detects SSO source profiles (looking up the source's section in the config parser) and routes to .sso(profileName: sourceProfile) instead of demanding static keys.
- New isSSOProfile(in:profile:) helper detects both modern sso_session and legacy sso_start_url formats.

This PR was produced with the assistance of Claude Code (Opus 4.7), but I have reviewed the entirety personally, and made changes where the LLM made bad decisions. I have tested this against a real project using Soto, and have run the test suite.

### DIFF
--- a/Sources/SotoCore/Credential/ConfigFileLoader.swift
+++ b/Sources/SotoCore/Credential/ConfigFileLoader.swift
@@ -176,7 +176,7 @@ enum ConfigFileLoader {
     ///   - fileIO: non-blocking file IO
     /// - Returns: INIParser
     static func loadINIFile(path: String, fileIO: NonBlockingFileIO) async throws -> INIParser {
-        let buffer: ByteBuffer 
+        let buffer: ByteBuffer
         do {
             buffer = try await loadFile(path: path, fileIO: fileIO)
         } catch let error as IOError where error.errnoCode == ENOENT {
--- a/Sources/SotoCore/Credential/ConfigFileLoader.swift
+++ b/Sources/SotoCore/Credential/ConfigFileLoader.swift
@@ -88,6 +88,7 @@ enum ConfigFileLoader {
     /// - missingSecretAccessKey: If the secret access key was not found
     public struct ConfigFileError: Error, Equatable {
         enum Internal: Equatable {
+            case fileDoesNotExist
             case invalidINIFile
             case missingProfile
             case missingAccessKeyId
@@ -95,6 +96,8 @@ enum ConfigFileLoader {
         }
         let value: Internal
 
+        /// Credentials or profile ini file does not exist
+        public static var fileDoesNotExist: Self { .init(value: .fileDoesNotExist) }
         /// Credentials or profile ini file failed to load
         public static var invalidINIFile: Self { .init(value: .invalidINIFile) }
         /// Cannot find profile in ini file
@@ -126,13 +129,13 @@ enum ConfigFileLoader {
         let credentialsINIParser: INIParser?
         do {
             credentialsINIParser = try await self.loadINIFile(path: credentialsFilePath, fileIO: fileIO)
-        } catch let error as IOError where error.errnoCode == ENOENT {
+        } catch let error as ConfigFileError where error == ConfigFileError.fileDoesNotExist {
             credentialsINIParser = nil
         }
         let configINIParser: INIParser?
         do {
             configINIParser = try await self.loadINIFile(path: configFilePath, fileIO: fileIO)
-        } catch let error as IOError where error.errnoCode == ENOENT {
+        } catch let error as ConfigFileError where error == ConfigFileError.fileDoesNotExist {
             configINIParser = nil
         }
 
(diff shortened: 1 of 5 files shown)


adam-fowler:
@sebsto any thoughts?

sebsto:
Thanks for the PR @nicksloan. I reviewed this against the AWS CLI v2 documentation and the current codebase (including my recent SSO commit in #686).

The `role_arn` + `source_profile` → SSO profile pattern is the standard multi-account layout produced by `aws configure sso`. For example:

```ini
[profile dev]
role_arn = arn:aws:iam::222222222222:role/Developer
source_profile = sso-base

[profile sso-base]
sso_session = my-org
sso_account_id = 111111111111
sso_role_name = SSOAccess

[sso-session my-org]
sso_start_url = https://my-org.awsapps.com/start
sso_region = us-east-1
```

The AWS CLI v2 resolves `source_profile` recursively through whatever credential mechanism the source uses (static, SSO, process, etc.). My #686 commit added SSO as a standalone provider in the chain but didn't handle SSO as the *source* for an assume-role chain. This PR fills that gap.

Ref: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html

**Concerns I'd like addressed:**

1. **Silent swallowing of credentials file errors** — Previously, any failure to load `~/.aws/credentials` threw `.noProvider`. Now *all* errors (including permission denied, corrupt file, etc.) are silently ignored. I'd prefer distinguishing "file not found" (tolerable, proceed with config-only) from "file exists but unreadable" (should still error). This avoids masking real misconfigurations.

2. **Partial SSO profile detection** — `isSSOProfile` checks for `sso_session` or `sso_start_url` but doesn't validate the profile is *complete* (has `sso_account_id`, `sso_role_name`). A partial config will be routed to `SSOCredentialProvider` which will fail with a less obvious error downstream. A validation or clearer error message would help.

3. **Test coverage** — Codecov flags 4 uncovered lines. Specifically, the path where the credentials file exists but the profile is missing from it AND the config supplies a `role_arn` (the `catch .missingProfile` branch) isn't tested.

Overall this is a solid contribution that addresses a common real-world configuration. Happy to discuss the error-handling concern further.

nicksloan:
@sebsto Thanks for the feedback. The test coverage was handled by my latest commit as shown in the latest Codecov run.

Your first point seems like a great improvement to me. I'll work on that.

On the second point, the behavior now is that we raise the following error:

```
2026-05-21T12:18:21-0400 error MyProject: aws-error-message=ssoConfigMissing: Profile 'nicksloan-admin' does not have SSO configuration. Configure with 'aws configure sso'. aws-operation=AssumeRole aws-request-id=1 aws-service=sts [SotoCore] AWSClient error
```

This is slightly more informative than the AWS CLI (2.34.24) error in the same situation, because it references the attempt at SSO:

```
aws: [ERROR]: An error occurred (Configuration): Unable to retrieve credentials: The source profile "nicksloan-admin" must have credentials.
```

The profile specified by AWS_PROFILE was `nicksloan-test`. It might be confusing in this instance because the error is pointing to a profile that is not the profile the user specified, but it is correctly pointing at the profile with the problem to be corrected.

sebsto:
Thank you Nick.

> Your first point seems like a great improvement to me. I'll work on that.

OK

> On the second point, the behavior now is that we raise the following error:

If it is simple to report the profile that was specified and the one where the error is, it would be a great DX. But it's a minor point. The existing error message is good for me.

nicksloan:
@sebsto I think the latest on this PR should address the first concern you raised, and the second commit attaches the origin profile to error messages that come from a credential chain.

Thanks for your feedback so far, and let me know if you have any further concerns.

sebsto:
Thank you Nick. Just one small question: should the two new struct be public ? Unless I missed something they are exposed to clients, aren’t they ?

nicksloan:
Good catch on the internal error. I've made that public. I'm not sure ProfileScopedCredentialProvider needs to be. It is initialized by ConfigFileCredentialProvider, and I'm not sure how valuable it is to use directly, but I'd be happy to make it public if you disagree.

sebsto (approved):
LGTM. Thank you

Sources/SotoCore/Credential/ConfigFileLoader.swift
```diff
@@ -120,28 +120,39 @@ enum ConfigFileLoader {
...
-            // Load credentials file
-            credentialsINIParser = try await self.loadINIFile(
-                path: credentialsFilePath,
-                fileIO: fileIO
-            )
-        } catch {
-            // Throw `.noProvider` error if credential file cannot be loaded
-            throw CredentialProviderError.noProvider
+            credentialsINIParser = try await self.loadINIFile(path: credentialsFilePath, fileIO: fileIO)
+        } catch let error as IOError where error.errnoCode == ENOENT {
```
adam-fowler: Can we catch this inside `loadINIFile`, and throw a (new) `ConfigFileError.fileDoesNotExist`

adam-fowler (commented):
I'm pretty happy with this, one minor thing and it can go in

adam-fowler (approved):

adam-fowler:
Looks like there is swift-format issue

nicksloan:
Trailing white space? Woof. That's fixed now.

Sources/SotoCore/Credential/ConfigFileLoader.swift
```diff
@@ -176,7 +176,7 @@ enum ConfigFileLoader {
     ///   - fileIO: non-blocking file IO
     /// - Returns: INIParser
     static func loadINIFile(path: String, fileIO: NonBlockingFileIO) async throws -> INIParser {
-        let buffer: ByteBuffer 
+        let buffer: ByteBuffer
```
adam-fowler, line 179: The important shit

PR #692, opened 2026-06-04
Make LoginCredentialProvider token persistence best-effort

### LINKED ISSUE #691: LoginCredentialProvider: token refresh fails in sandboxed environments when cache write is blocked
## Summary

`LoginCredentialProvider.getCredential()` treats a failed disk write of the refreshed token as a fatal error, discarding the perfectly valid in-memory credentials it just received from the refresh endpoint. In environments where writing to `~/.aws/login/cache/` is blocked (e.g. a SwiftPM command plugin sandbox), every invocation that requires a refresh fails with `No credential provider found`, even though the network refresh itself succeeded.

## How to reproduce

1. Create a SwiftPM command plugin that uses `AWSClient` with the default credential provider chain.
2. Configure AWS credentials via `aws login` (Identity Center / Isengard flow, cached at `~/.aws/login/cache/<sha256>.json`).
3. Wait for the cached credentials to expire (or set `expiresAt` to a past date in the cache file).
4. Run the plugin: `swift package --allow-network-connections all:443 <plugin-verb>`
5. Result: `No credential provider found`.
6. Run `aws sts get-caller-identity` from a normal shell (this refreshes the cache file outside the sandbox).
7. Re-run the plugin → succeeds, because no refresh is needed.

## Root cause

In `Sources/SotoCore/Credential/Login/LoginCredentialProvider.swift`, `getCredential()`:

1. Reads the token file ✅ (sandbox allows reads)
2. If still valid, returns it ✅
3. POSTs to the refresh endpoint with DPoP ✅ (sandbox allows outbound 443)
4. Receives fresh credentials ✅
5. Calls `tokenFileManager.saveToken(...)` ❌ — SwiftPM plugin sandbox blocks writes outside the package/build directory
6. The throw propagates out of `getCredential()`, the chain falls through to `NullCredentialProvider`, and the user sees `No credential provider found` even though step 4 already produced valid credentials.

The SwiftPM plugin sandbox on macOS allows reads anywhere, allows outbound 443 with `--allow-network-connections all:443`, but does not allow writes to `~/.aws/login/cache/`. The AWS CLI works because it runs outside the sandbox; SotoCore inherits the sandbox restrictions of the host process.

## Proposed fix

Make `saveToken` best-effort: log a warning on failure, but still return the freshly-issued in-memory credentials so the current session can proceed. The next invocation will re-refresh, which is the same behavior the user experiences today when the credentials genuinely expire.

A PR with the fix and a regression test is on the way.

sebsto:
Fixes #691.

## Summary

- `LoginCredentialProvider.getCredential()` was treating a failed disk write of the refreshed token as fatal, throwing away the perfectly valid credentials it just received from the refresh endpoint.
- In sandboxed environments (SwiftPM command plugins, App Sandbox containers, etc.) where reads from `~/.aws/login/cache/` are allowed but writes are blocked, every refresh path failed with `No credential provider found`.
- This change wraps `tokenFileManager.saveToken(...)` in a `do/catch`, logs at warning level, and returns the freshly-issued in-memory credentials regardless of whether the disk save succeeded. Next invocation will re-refresh — same behavior as today when the cached credentials legitimately expire.

## Test plan

- [x] Added `getCredentialsDespiteWriteFailure` regression test that:
  - Seeds an expired token in a temp cache directory.
  - `chmod 555`s the directory to simulate a sandbox blocking writes.
  - Runs the refresh against a mock HTTP client returning fresh credentials.
  - Asserts the fresh credentials are returned even though the on-disk save failed.
- [x] `swift test --filter LoginCredentialProviderTests` — all 10 tests pass (9 existing + 1 new).
- [x] `swift build` — clean (one pre-existing `NonBlockingFileIO` deprecation warning, unrelated to this change).

### DIFF
--- a/Sources/SotoCore/Credential/Login/LoginCredentialProvider.swift
+++ b/Sources/SotoCore/Credential/Login/LoginCredentialProvider.swift
@@ -206,10 +206,18 @@ public struct LoginCredentialProvider: CredentialProvider {
             refreshToken: tokenResponse.refreshToken
         )
 
-        // Save updated token to disk
-        try await tokenFileManager.saveToken(updatedToken, to: tokenPath, fileIO: fileIO, threadPool: threadPool)
+        // Save updated token to disk (best-effort — may fail in sandboxed environments
+        // such as SwiftPM plugins where writes outside the package directory are blocked)
+        do {
+            try await tokenFileManager.saveToken(updatedToken, to: tokenPath, fileIO: fileIO, threadPool: threadPool)
+        } catch {
+            logger.warning(
+                "Could not persist refreshed credentials to disk. Credentials are valid for this session but will need to be refreshed again next time.",
+                metadata: ["error": .string("\(error)")]
+            )
+        }
 
-        // Return expiring credential
+        // Return expiring credential regardless of whether disk save succeeded
         return RotatingCredential(
             accessKeyId: tokenResponse.accessToken.accessKeyId,
             secretAccessKey: tokenResponse.accessToken.secretAccessKey,
--- a/Tests/SotoCoreTests/Credential/Login/LoginCredentialProviderTests.swift
+++ b/Tests/SotoCoreTests/Credential/Login/LoginCredentialProviderTests.swift
@@ -511,6 +511,82 @@ final class LoginCredentialProviderTests {
         }
     }
 
+    @Test("Get credentials succeeds when token file write fails (sandbox)")
+    func getCredentialsDespiteWriteFailure() async throws {
+        let tempDirectory = try createTempDirectory()
+        defer {
+            // Restore permissions so cleanup can succeed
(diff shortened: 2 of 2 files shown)


sebsto:
The Linux CI failure is in `RotatingCredentialProviderTests.testAlwaysGetNewTokenIfTokenLifetimeForUseIsShorterThanLifetime` — a timing-dependent test in `RotatingCredentialProvider`, which this PR does not touch (the change is isolated to `LoginCredentialProvider`'s save-token error handling). macOS passed, only Linux Swift 6.3 failed (and 6.1/6.2 were cancelled by the matrix `fail-fast`). The test passes locally.

Looks like a flake — could you re-run the failed Linux jobs? I don't have permission to do it from my fork.

adam-fowler (approved):

PR #694, opened 2026-06-05
Remove platforms declaration, use @available with AvailabilityMacro

### LINKED ISSUE #693: Remove platforms: declaration from Package.swift, use @available instead
We depend on soto-core in [swift-aws-lambda-runtime](https://github.com/awslabs/swift-aws-lambda-runtime) for credential management and request signing in our deploy plugin.

## Problem

soto-core declares `platforms: [.macOS(.v10_15), ...]` in its Package.swift. This forces any downstream package that depends on soto-core to also declare a `platforms:` minimum, even if the downstream package prefers to use `@available` annotations instead.

Without a `platforms:` declaration, SwiftPM defaults to macOS 10.13. When an executable target depends on soto-core, SwiftPM rejects the build at planning time:

```
error: the executable "AWSLambdaPluginHelper" requires macos 10.13,
but depends on the product "SotoCore" which requires macos 10.15;
consider changing the executable to require macos 10.15 or later.
```

This is a SwiftPM-level check that happens before compilation. `@available` annotations cannot solve it because SwiftPM refuses to plan the build.

## What we had to do

We had to add `platforms: [.macOS(.v15)]` to our Package.swift, which we previously avoided. Our project uses a custom availability macro (`@available(LambdaSwift 2.0, *)` mapping to macOS 15.0) on all types instead of a package-wide platform declaration. Adding `platforms:` was the only workaround.

## Suggested fix

Remove the `platforms:` declaration from soto-core Package.swift and use `@available(macOS 10.15, *)` annotations on types/functions that require macOS 10.15+ APIs. This way downstream packages are not forced to declare their own platform minimum.

This is the same approach used by swift-nio and other server-side Swift packages that need to support both macOS (for development) and Linux (for deployment) without imposing platform constraints on dependents.

## Context

This issue was encountered while working on the swift-aws-lambda-runtime deploy plugin (awslabs/swift-aws-lambda-runtime PR 567).

sebsto:
## Summary

- Remove the `platforms:` declaration from `Package.swift` so downstream packages are not forced to declare their own platform minimum
- Add a custom `AvailabilityMacro` (`SotoCore 7.0` → macOS 10.15, iOS 13, tvOS 13, watchOS 6) to express the same constraint via `@available` annotations
- Annotate all types/extensions requiring macOS 10.15+ (async/await, CryptoKit, Sendable) with `@available(SotoCore 7.0, *)`
- Keep `platforms:` in `Package@swift-6.0.swift` for backwards compatibility with toolchains that don't support AvailabilityMacro

## Motivation

Closes #693

soto-core's `platforms:` declaration forces downstream packages to also declare a `platforms:` minimum, even if they prefer to use `@available` annotations instead. This caused issues for [swift-aws-lambda-runtime](https://github.com/awslabs/swift-aws-lambda-runtime/pull/567) which uses a custom availability macro approach.

This follows the same pattern used by [Hummingbird](https://github.com/hummingbird-project/hummingbird/blob/main/Package.swift) and other server-side Swift packages.

## Test plan

- [x] Builds cleanly with Swift 6.0 (`Package@swift-6.0.swift` with `platforms:`)
- [x] Builds cleanly with Swift 6.1 (`Package.swift` with AvailabilityMacro)
- [x] Builds cleanly with Swift 6.3 (`Package.swift` with AvailabilityMacro + Android)
- [x] All tests pass
- [x] Verified downstream `soto` project compiles correctly with this version


adam-fowler (commented):
Minor gripe. Can you stick the `@available` below the comment for a type. It is part of the type definition. Seems odd to add the comment in the middle. 

```swift
/// My thing does stuff
@available(Soto 7.0, *)
struct MyThing {}
```

sebsto:
Done

adam-fowler:
You need to require swift-crypto 3.12.3. Otherwise swift-crypto will include platform requirements

sebsto:
Done, I added 
```
        .package(url: "https://github.com/apple/swift-crypto.git", "3.12.3"..<"5.0.0"),
```

But this might cause some compatibility problems to packages using Soto Core and having strict (old) dep requirements on Swift Crypto, isn't it ?

sebsto:
@adam-fowler Should we close this ? (because of the required version bump for swift-crypto) or should we wait for Swift 6.4 ?

PR #702, opened 2026-06-23
Move some tests across to swift testing

adam-fowler:
Can't push everything across to swift-testing as quite a lot of the tests use resources from machine like files, environment variables, ports. We should probably fix the test server to use the Testing event loop groups.


Tests/SotoXMLTests/XMLTests.swift
```diff
@@ -12,123 +12,124 @@
...
-import XCTest
+import Foundation
+import Testing
 
 @testable import SotoXML
 
-class XMLTests: XCTestCase {
+class XMLTests {
     /// helper test function to use throughout all the decode/encode tests
     func testDecodeEncode(xml: String) throws {
```
0xTim, line 22: This looks like it's missing an `@Test`

0xTim (changes requested):
There's a missing `@Test`

One optional would be to add `@Suite` to the different test classes so they're separated rather than having a giant list of tests

Tests/SotoXMLTests/XMLTests.swift
adam-fowler: That's a helper function. Not an actual test.

0xTim: 🤦 comment even says it

0xTim (approved):

